PR #126, opened 2014-09-16
Routes.configure(...) at JS level

ksz2k:
Hi

I needed `Routes.set_prefix()` to `defaults['prefix']` be setable based on currently signed in user UID so here's little fix - hoping it'll be usefull for you too.. ;) 

So with this patch we can do the following:

``` javascript
Routes.set_prefix('/ada46feca');

Routes.root_path()
>> '/ada46feca/'
```

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -129,6 +129,18 @@ In order to make routes helpers available globally:
 jQuery.extend(window, Routes)
 ```
 
+## Configuring in javascript
+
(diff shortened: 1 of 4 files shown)


Galathius:
👍

le0pard:
@bogdan looks like we need to close it.

bogdan:
I think not. I keep as a reminder for a good feature that we don't have.

bogdan:
Implemented in master

PR #142, opened 2015-03-05
Working with custom routes file

FabioMR:
Sometimes I need to change the default routes path or define multiples files. So, I've adapted the code to this situation.

### DIFF
--- a/lib/js_routes/engine.rb
+++ b/lib/js_routes/engine.rb
@@ -4,10 +4,17 @@ class JsRoutes
 
     initializer 'js-routes.dependent_on_routes', after: "sprockets.environment" do
       if Rails.application.assets.respond_to?(:register_preprocessor)
-        routes = Rails.root.join('config','routes.rb')
-        Rails.application.assets.register_preprocessor 'application/javascript', :'js-routes_dependent_on_routes' do |ctx,data|
-          ctx.depend_on(routes) if ctx.logical_path == JS_ROUTES_ASSET
-          data
+        if Rails::VERSION::MAJOR == 3
+          route_paths = Rails.application.paths['config/routes'].to_a
+        else
+          route_paths = Rails.application.paths['config/routes.rb'].to_a
+        end
+
+        route_paths.each do |path|
+          Rails.application.assets.register_preprocessor 'application/javascript', :'js-routes_dependent_on_routes' do |ctx,data|
+            ctx.depend_on(Pathname.new(path)) if ctx.logical_path == JS_ROUTES_ASSET
+            data
+          end
         end
       end
     end
--- a/lib/js_routes/engine.rb
+++ b/lib/js_routes/engine.rb
@@ -4,11 +4,7 @@ class JsRoutes
 
     initializer 'js-routes.dependent_on_routes', after: "sprockets.environment" do
       if Rails.application.assets.respond_to?(:register_preprocessor)
-        if Rails::VERSION::MAJOR == 3
-          route_paths = Rails.application.paths['config/routes'].to_a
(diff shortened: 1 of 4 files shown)


le0pard:
Hello @FabioMR . To modify JS you should edit https://github.com/railsware/js-routes/blob/master/lib/routes.js.coffee instead of routes.js

bogdan:
I didn't find any documentation on using `Rails.application.paths`. Can some suggest me a link?

FabioMR:
@le0pard In fact, I didn't modified the routes.js. I just ran rspec and It was builded automatically based on .coffee file.

FabioMR:
@bogdan You are already using Rails.application.paths in your code: https://github.com/railsware/js-routes/blob/master/spec/spec_helper.rb
I used the same idea, just changed the code to works for my context.

By the way, this is the guy: http://api.rubyonrails.org/classes/Rails/Paths/Path.html

le0pard:
@FabioMR Ok, it is just new coffescript gem.

FabioMR:
@le0pard Ok. But any problem? If you want, I can back routes.js changes.

le0pard:
@bogdan will review PR, I just checked pull request on errors.

lib/js_routes/engine.rb
```diff
@@ -4,10 +4,17 @@ class Engine < ::Rails::Engine
...
       if Rails.application.assets.respond_to?(:register_preprocessor)
-        routes = Rails.root.join('config','routes.rb')
-        Rails.application.assets.register_preprocessor 'application/javascript', :'js-routes_dependent_on_routes' do |ctx,data|
-          ctx.depend_on(routes) if ctx.logical_path == JS_ROUTES_ASSET
-          data
+        if Rails::VERSION::MAJOR == 3
+          route_paths = Rails.application.paths['config/routes'].to_a
+        else
+          route_paths = Rails.application.paths['config/routes.rb'].to_a
+        end
```
le0pard: `route_paths = Rails.application.paths["config/routes#{3 == Rails::VERSION::MAJOR ? '' : '.rb'}"].to_a`

FabioMR: Ok, less messy! I'll change it!

bogdan:
I don't see any doc on how `paths` should be used. That's why it looks strange to me that you are using `paths` in an undocumented way (maybe along with js-routes test suite - but it is ok there because test suite runtime environment is a hack). 

Maybe this patch makes it work for your case and break it for other person case. 
The `Paths::Path` didn't give an idea how I can use `Rails.application.paths`.  It would be cool to see Rails doc or how-to of how the `paths` are used.

FabioMR:
@bogdan I can't see how this patch can break in some cases, because it is how Rails use the routes. But I agree that the doc is very poor. I think I get your point.

But how can we go with this? I could change the route path through a config, what do your think? The config could have a default value and we can change it if necessary.

bogdan:
Here is what I've found: https://github.com/rails/rails/commit/5e7d6bba79393de0279917f93b82f3b7b176f4b5
https://github.com/rails/rails/compare/fa736e69a197...2d9dbf416b14

It looks like this feature works only in 3.x.x series and you are forced to use concerns in 4.x.x

FabioMR:
I tested my project with the original config/routes.rb and just imported other files. It worked fine too. So I think this pull request can be closed, it is not necessary anymore!

I hope this issue could help others with the same problem as I.

Thank you!

**Solution:**
`#config/routes.rb`
`require_relative 'routes/application.rb'`
`require_relative 'routes/api.rb'`

PR #143, opened 2015-04-13
fix for sprockets-3.0.x

masarakki:
after sprockets-3.0.x, I can't find depend_on files when assets.paths dosen't include directory.


le0pard:
Hello, @masarakki . Thanks for pull request. But also need fix tests, because https://github.com/rails/sprockets/blob/v3.0.0/lib/sprockets/context.rb also changed in sprockets 3.0.

masarakki:
How can I write spec for both 3.0.x and 2.x? use Appraisal?

le0pard:
Yes, you can use Appraisal and also you can write conditions inside specs (you have access to `Sprockets::VERSION`, like `Sprockets::VERSION =~ /\A3\.\d+\.\d+$/`)

le0pard:
Thanks @masarakki, this pull request looks like good. Need review from @bogdan (I think better remove 1.9.3 from CI matrix - it is already unsupported).

lib/js_routes/engine.rb
```diff
@@ -2,9 +2,13 @@ class JsRoutes
   class Engine < ::Rails::Engine
     JS_ROUTES_ASSET = 'js-routes'
 
+    initializer 'js-routes.append_assets_path', group: :all, after: :append_assets_path do |app|
+      app.config.assets.paths.unshift Rails.root.join('config').to_s
```
bogdan: Why do we need to add `config` to assets path?

masarakki:
`depend_on` of Sprockets 3.0.x check the requested file is under the `assets_paths`

https://github.com/rails/sprockets/blob/master/lib/sprockets/resolve.rb#L69

lib/js_routes/engine.rb
le0pard: https://github.com/railsware/js-routes/pull/143#issuecomment-92674049

le0pard:
Hello @masarakki . 

I think we don't need add

```ruby
initializer 'js-routes.append_assets_path', group: :all, after: :append_assets_path do |app|
  app.config.assets.paths.unshift Rails.root.join('config').to_s
end
```
to assets_paths, because we render js-routes from https://github.com/railsware/js-routes/blob/master/app/assets/javascripts/js-routes.js.erb file, which is located in assets_path (I hope). Does it work without this?

pbhogan:
It is technically necessary since sprockets checks that all dependencies are themselves under the asset paths and routes.rb is not. However, adding config to the asset paths is not good in my opinion. It may have other unwanted side effects.

A mechanism seems to exist to declare external environmental dependencies in sprockets 3. I've taken a shot at implementing using it myself and this seems to work:
https://github.com/pbhogan/js-routes/blob/master/lib/js_routes/engine.rb

I've left the original `register_preprocessor` call in since it is necessary for sprockets 2 but is just rejected by sprockets 3 harmlessly in `depend_on` here:
https://github.com/sstephenson/sprockets/blob/v3.0.0.beta.8/lib/sprockets/resolve.rb#L65

masarakki:
@pbhogan 
it seems good!!

masarakki:
update code.

lib/js_routes/engine.rb
```diff
@@ -3,13 +3,19 @@ class Engine < ::Rails::Engine
...
-        routes = Rails.root.join('config','routes.rb')
         Rails.application.assets.register_preprocessor 'application/javascript', :'js-routes_dependent_on_routes' do |ctx,data|
           ctx.depend_on(routes) if ctx.logical_path == JS_ROUTES_ASSET
           data
         end
       end
+
+      # only sprockets >= 3.0
+      if  Rails.application.assets.respond_to?(:depend_on)
+        Rails.application.assets.depend_on "file-digetst://#{routes}"
```
le0pard: Little type: `file-digest://` should be.

masarakki:
oh, thx!

le0pard:
waiting @bogdan

spec/js_routes/zzz_last_post_rails_init_spec.rb
```diff
@@ -6,17 +6,42 @@
...
     JsRoutes.generate!(NAME)
-    Rails.configuration.eager_load = false
-    Rails.application.initialize!
   end
 
-  after(:all) do
-    FileUtils.rm_f(NAME)
+  after(:each) do
+    FileUtils.rm_rf Rails.root.join('tmp/cache')
+    FileUtils.rm_f NAME
```
bogdan: I think remove those files in `before` is enough. Why do you think `after` makes sense?

bogdan:
Looks good in general. Only one small comment regarding rspec callbacks.

le0pard:
ping @bogdan :)

PR #149, opened 2015-06-16
Use flat_map rather than map.flatten

glittershark:
This is a small performance and readability boost.


le0pard:
Agree - http://gistflow.com/posts/578-use-flat_map-instead-of-map-flatten @glittershark, can you add this to CHANGELOG.md?

glittershark:
Sure! Would you rather that in a separate commit or squashed into one?

le0pard:
"squashed into one" please :+1:

glittershark:
done!

le0pard:
Thanks!

PR #150, opened 2015-06-17
URL escape routes.rb url to fix bad `URI(is not URI?)` error

kieranklaassen:
This fixes a bug that that crashed the Gem when you have a space in your path.


reubenbrown:
+1

pelletencate:
+1
Holy sh*t, this is awesome, @kieranklaassen you're a hero! I never understood what the problem was, I realised it didn't work with spaces in my path so I changed my entire directory tree using underscores instead. But it still scares me when I put my stuff to a PaaS, you never know what they do behind the scenes.

I can't wait for this fix to be released.

le0pard:
Hello, @kieranklaassen . Please, add CHANGELOG.md you change.

kieranklaassen:
@le0pard Done sir!

aksh0925:
+1

PR #151, opened 2015-06-18
fix for rails 5 - test rails-edge on travis allowing failure

twalpole:
Fix for rails 5 removal of Route#optional_parts.
Adds rails-edge to appraisal and tests on travis - allowing failure


bogdan:
It is very weird that `optional_parts` got removed. Are you sure it wasn't renamed?
Please give a link to commit that removes it.

We need to consider why did they remove it. Maybe they think there should be no optional parts any more or other crazy decision.

twalpole:
It was removed here.  https://github.com/rails/rails/commit/198b1dd201f7fff5aad696273f731c0b204996d5

le0pard:
@bogdan https://github.com/rails/rails/pull/17930

bogdan:
Lets copy the implementation that is there than. `path.optional_names.map(&:to_sym)`. It feels more correct. Not sure if we need `to_sym`.

twalpole:
It feels more correct to me to stay at the higher level of Route rather than drop down into the implementation of path.  It's more isolated from any future changes

bogdan:
From the other hand you've assumed that there is no other routes than required and optional. `Not required route is always optional` might not be true. It suppose to be in any adequate gem. But rails is not so adequate.

twalpole:
Agreed, but since Path#required_names is implemented as names-optional_names it is currently the case, and if that ever changes there are bigger problems.  I feel the current proposed fix is easy to read, easy to understand, and decently isolated from underlying implementation changes

bogdan:
Ok, not a big deal to change anyway.

PR #154, opened 2015-06-25
#152 - prefix generated files with comment

### LINKED ISSUE #152: Adding a comment to the top of generated files
For those of us who work on larger dev teams, it would be nice to have a comment at the top of any generated routes.js files—something which just explains that it _is_ a generated file and maybe a hint about how it is generated—so that other (and future) devs know where the file came from and also to warn anyone from trying to modify it by hand.

Perhaps this could be added as an option, so that it would not be mandatory. It would be really great to be able to do something like this:

```
config.comment = "This file was generated by the js-routes gem."
```

Would there be any interest in something like this if I put together a pull request?

jordanstephens:
Fixes #152 

Prefix generated files with a comment to indicate that the files are generated by the js-routes gem based on Rails routes for the given application.


lib/routes.js
```diff
@@ -1,3 +1,9 @@
+
```
bogdan: Please remove this blank line.

bogdan: This is important for JS compressors that don't delete a first comment aka banner in a file and that blank line may cause them to assume there is no banner.

jordanstephens: @bogdan Curiously, if I remove that line and run the specs, the CoffeeScript compiler continues to output that file with the blank line at the top. If you'd like, I suppose we could pass the output through `String#strip` here: https://github.com/railsware/js-routes/blob/master/spec/spec_helper.rb#L138

bogdan: Yes, lets do `strip` in case it will not delete line break at the last line or `lstrip` otherwise.

jordanstephens: Alright, added `lstrip` to the spec helper code.

Issue #153, opened 2015-06-24
Removing URL params with undefined or null values


jordanstephens:
I'm using this library all over the place in a very large application, and most of those places where I'm passing state objects as params to `Route.*_path()` functions. There are many cases where some of object properties are `undefined` or `null` at various times ending up with query strings which look like this:

    ?a=&b=&foo=bar

It would be really great if these functions automatically filtered out `null` or `undefined` values before building the query strings.

The nature of this library is that there are a ton of different, but very similar functions which end up getting called all over the place. Because of this, there isn't really a path to fix this kind of issue in my application code without writing code to filter these values in every single place we call one of these functions.

Surely the vast majority of users are not interested in params with empty values ending up in url query strings, especially, when you are working with many keys which may potentially be empty.

I will be happy to submit a pull request to have this fixed if you will merge it.

Thanks

bogdan:
js-routes's serialization was made compatible with ones provided by jQuery and Rails.

``` 
jQuery.param({a: undefined, b: null, c: false}) // => "a=&b=&c=false"
{a: nil, b: false}.to_query # => "a=&b=false"
```

We don't aim for fixing word wide accepted spec here.

bogdan:
The thing I can propose is make `serialize` part of the API so that you can redefine it. Wdyt?

jordanstephens:
You are totally right that `jQuery.param` and  `ActiveSupport::CoreExtensions::Hash#to_query` work this way, but I don't think that either of those are the correct model for comparison to the route helpers provided by js-routes.

It seems like one of the primary purposes of js-routes is to provide javascript route helpers with the same functionality as the "named route helpers" in Rails.

Let's assume we have the following Rails route definition:

    get "/users/:id", to: "users#show", as: :user

Rails gives us this nice `user_path` helper:

    user_path(1) # => "/users/1"

which corresponds nicely to the following `Routes.user_path` helper provided by js-routes:

    Routes.user_path(1) // => "/users/1"

This is great! It's really cool to have route helpers that work the same way on both the server and the client side of my application. 

Now let's take a look at what happens when I use the same helpers with parameters which may be  `nil` or `null`:

Rails actually excludes keys with `nil` values from the query string:

    user_path(1, { foo: "bar", baz: nil }) # => "/users/1?foo=bar"

while js-routes leaves the key in the string and gives it an empty value:

    Routes.user_path(1, { foo: "bar", baz: null }) // => "/users/1?foo=bar&baz="

[...]

PR #155, opened 2015-06-30
Allow Configurable Serializer Function (Solves #153)

jordanstephens:
Here's a first pass on a configurable serializer for issue #153.

Usage is as described in that issue.

```ruby
JsRoutes.setup do |config|
  config.serializer = "myApp.myCustomSerializerFn"
end
```

Also note that I removed a bunch of "if jQuery present" specs as they should no longer be relevant.

If everything looks good to you guys, I'll update the readme.

Let me know if you guys have any questions.

Thanks!

### DIFF
--- a/lib/js_routes.rb
+++ b/lib/js_routes.rb
@@ -19,7 +19,8 @@ class JsRoutes
     url_links: nil,
     camel_case: false,
     default_url_options: {},
-    compact: false
+    compact: false,
+    serializer: nil
   }
 
   NODE_TYPES = {
@@ -104,6 +105,7 @@ class JsRoutes
     js.gsub!("DEFAULT_URL_OPTIONS", json(@options[:default_url_options].merge(deprecated_default_format)))
     js.gsub!("PREFIX", @options[:prefix] || "")
     js.gsub!("NODE_TYPES", json(NODE_TYPES))
+    js.gsub!("SERIALIZER", @options[:serializer] || "null")
     js.gsub!("ROUTES", js_routes)
   end
 
--- a/spec/js_routes/options_spec.rb
+++ b/spec/js_routes/options_spec.rb
@@ -13,6 +13,21 @@ describe JsRoutes, "options" do
   let(:_options) { {} }
   let(:_warnings) { true }
 
+  context "when serializer is specified" do
+    let(:_options) { {:serializer => "myCustomSerializer"} }
+
+    it "should set configurable serializer" do
+      # define custom serializer
+      # this is a nonsense serializer, which always returns foo=bar
+      # for all inputs
+      evaljs(%q(function myCustomSerializer(object, prefix) { return "foo=bar"; }))
+
+      # expect the nonsense serializer above to have appened foo=bar
+      # to the end of the path
+      expect(evaljs(%q(Routes.inboxes_path()))).to eql("/inboxes?foo=bar")
+    end
+  end
+
   context "when exclude is specified" do
 
     let(:_options) { {:exclude => /^admin_/} }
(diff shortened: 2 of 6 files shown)


lib/js_routes.rb
```diff
@@ -19,7 +19,8 @@ class JsRoutes
     url_links: nil,
     camel_case: false,
     default_url_options: {},
-    compact: false
+    compact: false,
+    serializer: "this.default_serializer"
```
bogdan: I think we should use nil here. This value feels too magical if someone will try to debug it.

spec/js_routes/options_spec.rb
```diff
@@ -13,6 +13,21 @@
...
+
+    it "should set configurable serializer" do
+      # define custom serializer
+      # this is a nonsense serializer, which always returns foo=bar
+      # for all inputs
+      evaljs(%q(function myCustomSerializer(object, prefix) { return "foo=bar"; }))
+
+      # expect the nonsense serializer above to have appened foo=bar
+      # to the end of the path
+      expect(evaljs(%q(Routes.inboxes_path()))).to include("?foo=bar")
```
bogdan: let's make exact match instead of equal.

lib/js_routes.rb
jordanstephens: @bogdan If we set the default to `nil` here, we will have to do a check somewhere else to set the default initializer. It has to get set somewhere, we can't just have a `nil` initializer floating around. Where would you prefer it get set?

Something like this on line 108?

    js.gsub!("SERIALIZER", @options[:serializer] || "this.default_serializer")

I think setting in the defaults is better than this option. This feels like a magic value. And really, `this.default_serializer` _is_ the default—not `nil`. It doesn't make sense to have a `nil` initializer.

spec/js_routes/options_spec.rb
jordanstephens: @bogdan I did this because this particular test isn't concerned with the rest of the output of `Routes.inboxes_path` the `/inboxes` part should be tested by some other test somewhere else. This test is only concerned with the `?foo=bar` portion of the output.

If you really prefer it, I can switch the test to the code below, but I wrote it the other way on purpose.

    expect(evaljs(%q(Routes.inboxes_path()))).to eql("/inboxes?foo=bar")

bogdan: Yes, I really prefer. We had bugs before when output had a wrong order e.g anchor was before query string.

lib/js_routes.rb
bogdan: The way we can make it clear is to make `default_serializer` a public method and let default be `JsRoutes.default_serialize`. In this way people can clearly see where is the default and even test in console it in console comparing to other serializers.

jordanstephens: @bogdan I'm not totally sure I follow. How does this work with the configurable namespace? We can't set the default to a public method on `JsRoutes` because we don't know that `JsRoutes` exists.

It might be better to just make `default_serializer` public when we run `createGlobalJsRoutesObject`, and then just add a README note that it is public on the namespace that the user defines, rather than try to get the default value to include the name of the configurable namespace.

spec/js_routes/options_spec.rb
jordanstephens: updated this to

    expect(evaljs(%q(Routes.inboxes_path()))).to eql("/inboxes?foo=bar")

lib/js_routes.rb
bogdan: Hm, you are right. namespace config value makes it not static. Let me think till tomorrow.

le0pard:
```ruby
js.gsub!("SERIALIZER", @options[:serializer] || "null")
```

and in JS:

```coffee
serializer = SERIALIZER
unless serializer?
  #use default
```

bogdan: I still vote for `nil` as default and implementation as @le0pard suggested.

jordanstephens:
@le0pard @bogdan alright, I set the default to `nil`, and moved the override logic into the javascript.

I think you guys are making the wrong call setting the default value to be a nonsense `nil` value rather than the actual default, and then hiding the override logic in the javascript. I don't see the benefit to the increased complexity in doing it this way. However, as it's your project, I'll follow your lead here.

I hope we can get this merged in soon.

Thanks

le0pard:
Funny, @jordanstephens . You think line "this.default_serializer" should mean something for developers, but one condition can slowdown all codebase :)

I think to not set initializer each time in function variable with serializer, we can cash it in `Utils` in this way:

```coffee
custom_serializer: SERIALIZER

serialize: (object) ->
   if @custom_serializer? and @get_object_type(@custom_serializer) is "function"
    @custom_serializer(object)
  else
    @default_serializer(object)
```

PR #159, opened 2015-09-01
Add window.location fallback for missing host parts

### LINKED ISSUE #157: Make `default_url_options.host` optional for *_url helpers
Problem: We have a multi-tenant app, where each customer has its own subdomain. We need `*_url` helpers to generate full URLs in various places. `default_url_options.host` only allows us to set one host, which will not work with subdomains.

Thoughts on making `host` optional and calculating `*_url` helper strings on the JavaScript side based on `default_url_options` and `*_path` helpers instead of entirely in Ruby?

andrewhavens:
### Added:
* Adds support for using the window.location.protocol when a protocol has not been specified.
* Adds support for using the window.location.host when a host has not been specified.
* Adds support for using the window.location.port when a port has not been specified.

Fixes #157.

@bogdan and @le0pard please review.

### DIFF
--- a/spec/js_routes/options_spec.rb
+++ b/spec/js_routes/options_spec.rb
@@ -308,12 +308,6 @@ describe JsRoutes, "options" do
     end
 
     context "when configuring with default_url_options" do
-      context "when default host is not specified" do
-        it "raises an error" do
-          expect { JsRoutes.generate({ :url_links => true }) }.to raise_error RuntimeError
-        end
-      end
-
       context "when only host option is specified" do
         let(:_options) { { :url_links => true, :default_url_options => {:host => "example.com"} } }
 
@@ -401,6 +395,42 @@ describe JsRoutes, "options" do
           expect(evaljs("Routes.new_session_url()")).to eq("https://example.com#{routes.new_session_path}")
           expect(evaljs("Routes.sso_url()")).to eq(routes.sso_url)
           expect(evaljs("Routes.portals_url()")).to eq("http://example.com:8080#{routes.portals_path}")
+        end
+      end
+    end
+
+    context 'when window.location is present' do
+      let(:current_protocol) { 'http:' } # window.location.protocol includes the colon character
+      let(:current_hostname) { 'current.example.com' }
+      let(:current_port){ '' } # an empty string means port 80
+      let(:current_host) do
+        host = "#{current_hostname}"
+        host += ":#{current_port}" unless current_port == ''
+        host
+      end
+
+      before do
+        jscontext['window'] = {
+          :location => {
+            :protocol => current_protocol,
+            :hostname => current_hostname,
+            :port => current_port,
+            :host => current_host
+          }
+        }
+      end
+
+      context "without specifying a default host" do
+        let(:_options) { { :url_links => true } }
+
+        it "uses the current host" do
+          expect { JsRoutes.generate(_options) }.not_to raise_error
+          expect(evaljs("Routes.inbox_path")).not_to be_nil
+          expect(evaljs("Routes.inbox_url")).not_to be_nil
+          expect(evaljs("Routes.inbox_url(1)")).to eq("http://current.example.com#{routes.inbox_path(1)}")
+          expect(evaljs("Routes.inbox_url(1, { test_key: \"test_val\" })")).to eq("http://current.example.com#{routes.inbox_path(1, :test_key => "test_val")}")
+          expect(evaljs("Routes.new_session_url()")).to eq("https://current.example.com#{routes.new_session_path}")
+          expect(evaljs("Routes.sso_url()")).to eq("http://sso.example.com#{routes.sso_path}")
         end
       end
     end
--- a/Readme.md
+++ b/Readme.md
@@ -65,7 +65,7 @@ Available options:
 * `url_links` (version >= 0.8.9) - Generate `*_url` helpers (in addition to the default `*_path` helpers).
   * Example: true
   * Default: false
-  * Be sure to specify a default host in `default_url_options`. Routes which specify a specific host, protocol, or port will be used instead of their corresponding default.
+  * Note: generated URLs will first use the protocol, host, and port options specified in the route definition. Otherwise, the URL will be based on the option specified in the `default_url_options` config. If no default option has been set, then the URL will fallback to the current URL based on `window.location`.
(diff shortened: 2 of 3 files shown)


spec/js_routes/options_spec.rb
```diff
@@ -308,12 +308,6 @@
     end
 
     context "when configuring with default_url_options" do
-      context "when default host is not specified" do
-        it "raises an error" do
-          expect { JsRoutes.generate({ :url_links => true }) }.to raise_error RuntimeError
```
andrewhavens: Note that we no longer raise an error when a default host hasn't been defined because we fallback to `window.location.host` if it has been specified.

bogdan: Yes, that's clear.

spec/js_routes/options_spec.rb
```diff
@@ -404,6 +398,42 @@
...
+        }
+      end
+
+      context "without specifying a default host" do
+        let(:_options) { { :url_links => true } }
+
+        it "uses the current host" do
+          expect { JsRoutes.generate(_options) }.not_to raise_error
+          expect(evaljs("Routes.inbox_path")).not_to be_nil
+          expect(evaljs("Routes.inbox_url")).not_to be_nil
```
bogdan: I don't know why these 3 assertions above make sense. Are we just fine having ones below?

bogdan:
The spec looks fine, but implementation is not that good for several reasons:

* we are generate too much code per route `_url` function. It should be as small as possible. My app has 337 routes. I imagine production apps going 1000+ easily. There routes js file will now take tons of space.

* We are generating too much JS via ruby. This is also very sad to mix 2 programming languages. We should only pass from ruby to JS and make all preparing manipulations only on either side of the interaction.

brauliobo:
+1

ianlotinsky:
@bogdan: while I don't necessarily disagree with your ideas for improvement, it seems like your concerns would require re-architecting other aspects of js-routes to address. Do you agree?

Is there measurable harm in releasing this first implementation?

le0pard:
I agreed with @bogdan - size of generated route is very important. Also right now not in all places check `window.location` object (this object is not present on Node.js). Right now is popular universal way of app (the same file have valid code for server and client). This https://github.com/railsware/js-routes/pull/159/files#diff-d9cb46b6cc4f05321735b6b24fdc04c6R228 and https://github.com/railsware/js-routes/pull/159/files#diff-d9cb46b6cc4f05321735b6b24fdc04c6R230can generate invalid js file for Node.js (and even with checks it still have invalid `_url` helpers for backend). But adding https://github.com/railsware/js-routes/pull/159/files#diff-d9cb46b6cc4f05321735b6b24fdc04c6R236 inside each route function will be bad for size of generated file.

I think most simple way, is to make `default_url_options` pass inside host/port JS function (like we doing with `serializer`). In this case you can do this in such way (I think this will not work right now):

```ruby
JsRoutes.setup do |config|
  config.default_url_options = {
    host: "function(){ return window.location.hostname; }()",
    port: "function(){ return window.location.port; }()"
  }
end
```

Or

```ruby
# some js code

function locationObject() {
  return window.location;
}

# ruby

JsRoutes.setup do |config|
  config.default_url_options = {
    host: "locationObject().hostname",
    port: "locationObject().port"
  }
end
```

I this case it is problem of developer how big will be generated js file and how valid it will be for Javascript environments this functions.

bogdan:
You can measure harm by seeing how much does it increase the routes files in currently released version and your branch. In this way we will know how much we will hurt production apps.

When the `*_url` feature was initially released, there was already a problem of generating too much JS, but it is not as near as in current one. So we will make evil to current users of `*_url` feature that do not need this new functionality.

Also I wouldn't say "reachitect" here. Maybe do a little of refactoring ~ 50 lines not more.
I am fine doing it myself this weekend.... and I'll probably do because it starts to eat my time anyway.

andrewhavens:
Sorry, I've been silent. I have been really busy at work for the past month and haven't had any free time to revisit this pull request.

I agree with @bogdan's and @le0pard's assessment. I think it's wasteful to calculate `window.location...` in every route. I wanted to create a helper function in JS that I could call instead. However, at the time that I created this, I did not know where to put it.

I might have a chance to work on this again next week. But @bogdan if you have a chance this weekend, feel free to pick up where I stopped.

bogdan:
Refactored code and merged. Tests left as original written in this PR. So there should be no surprises for you, @andrewhavens. 

But I have one question: I've found a little bit strange test written before https://github.com/railsware/js-routes/pull/137/files#diff-113171e2716568c39ab7bb4251c97260R344.

``` ruby
get '/' => 'sso#login', host: 'sso.example.com', as: :sso

let(:_options) { { :url_links => true, :default_url_options => {:host => "example.com", :port => 3000} } }


it "does not override host, protocol, or port when host is specified in route"         
  expect(evaljs("Routes.sso_url()")).to eq(routes.sso_url)
end 
```

This is really really weird thing. At first  you only check against host specified in route but not protocol and port. Maybe that is why this behaviour is not in place.

At second, I understand the reason of such behaviour, but rails doesn't do it and I don't know why js-routes should. E.g.:

``` ruby
site_url(1, protocol: 'ftp')  # => "ftp://localhost:3000/sites/1"
site_url(1, host: 'example.com')  # => "http://example.com:3000/sites/1"
```

So rails doesn't apply any magic like "Ignore port or protocol when host is specified" etc. in spite of it seems reasonable to me.

andrewhavens:
@bogdan It does seem confusing. I think the original purpose of that test was to make sure that if you specify a host in your route definition: 

    get '/' => 'sso#login', host: 'sso.example.com', as: :sso

...that it still uses that host, even if you specify a host is the js-routes default url options:

    let(:_options) { { :url_links => true, :default_url_options => {:host => "example.com", :port => 3000} } }

    it "does not override host, protocol, or port when host is specified in route"         
      expect(evaljs("Routes.sso_url()")).to eq('http://sos.example.com/')
    end 

Protocol wasn't mentioned in the default URL options, but I think if you were to specify a default protocol, the "sso" route would stay the same because it defines a specific host within the route definition.

bogdan:
I've put all the changes to make explicitly specified host, port and protocol parameters 99% compatible to rails. All changes are in master. @andrewhavens it would be cool to test in your project.

PR #160, opened 2015-09-08
Remove unnecessary assign

krzysiek1507:


le0pard:
Hello, @krzysiek1507 Thanks for PR, but js file compile from coffescript file: https://github.com/railsware/js-routes/blob/master/lib/routes.js.coffee#L18

krzysiek1507:
@le0pard, ok, thanks! ;)

PR #161, opened 2015-09-08
Serialize object with empty string value

krzysiek1507:


lib/routes.js.coffee
```diff
@@ -34,7 +34,7 @@ Utils =
           s.push "#{encodeURIComponent(prefix.toString())}=#{encodeURIComponent(object.toString())}"
 
     return "" unless s.length
-    s.join("&")
+    (item for item in s when item != "").join("&")
```
le0pard: `(item for item in s when item isnt "")`

krzysiek1507: ```Because the == operator frequently causes undesirable coercion, is intransitive, and has a different meaning than in other languages, CoffeeScript compiles == into ===, and != into !==. In addition, is compiles into ===, and isnt into !==.```

http://coffeescript.org/#operators

le0pard: I mean, we already use `isnt` in all file - better to be consistent.

krzysiek1507: @le0pard changed

le0pard:
@krzysiek1507 Also I think we need add test case to rails compatibility spec: https://github.com/railsware/js-routes/blob/master/spec/js_routes/rails_routes_compatibility_spec.rb

krzysiek1507:
@le0pard something like this?

le0pard:
@krzysiek1507 rails route should call with the same argument (what is why it is "compatibility" tests)

krzysiek1507:
With this changes:
```
Failure/Error: expect(evaljs("Routes.inboxes_path({a: '', b: 1})")).to eq(routes.inboxes_path(:a => '', :b => 1))

       expected: "/inboxes?a=&b=1"
            got: "/inboxes?b=1"
```

Without:
```
Failure/Error: expect(evaljs("Routes.inboxes_path({a: '', b: 1})")).to eq(routes.inboxes_path(:a => '', :b => 1))

       expected: "/inboxes?a=&b=1"
            got: "/inboxes?&b=1"
```

le0pard:
@krzysiek1507 perfect! Let's add this to tests

krzysiek1507:
@le0pard as pending?

le0pard:
Oh, sorry. So, this change broke rails compatibility. It is bad, because we should work in the same way in JS as rails route helpers. In this case better use custom serializer by option.

krzysiek1507:
Ok, but rails compatibility is broken now too. Second example is from master.

le0pard:
Ok, so first of all need fix serializer to leave `a=` for blank values

krzysiek1507:
@le0pard ok, fixed

le0pard:
cc @bogdan

le0pard:
Thanks @bogdan and @krzysiek1507 !

krzysiek1507:
Thanks!

Issue #168, opened 2015-11-09
tmp:cache:clear requirement slows down deployments


steakchaser:
Thanks for the nice gem, but having to call `tmp:cache:clear` is really slowing down asset compilation (local or server).  I realize the gem uses `Rails.application.routes` to inspect the routes, but is there any other way to do this?  Any ideas on how to speed things up?

bogdan:
How did you end up having it in your deployment?
You only need `tmp:cache:clear` when you made changes in js-routes configuration and you only need it in development. In production it wasn't ever an issue.

steakchaser:
We precompile assets on the server using the standard Capistrano 3 tasks, so in our case we have to clear the cache on the server.  

To work around it, we changed to a local precompile.

le0pard:
By default capistrano use `rake assets:precompile`, which recompile only changed files. So in this case no need to cleanup cache for each deploy.

bogdan:
I am thinking about this line of code:

https://github.com/railsware/js-routes/blob/master/lib/js_routes/engine.rb#L17

``` ruby
      routes = Rails.root.join('config', 'routes.rb').to_s
     ...
      Rails.application.assets.depend_on Rack::Utils.escape("file-digest://#{routes}")
```

We can have the same for `config/initializers/jsroutes.rb`. If I got it correctly, it will help sprockets to know that routes.js should be recompiled on changes at this initialisation file.

steakchaser:
@bogdan are you suggesting that the above would tell sprockets to re-compile js-routes if the actual `routes.rb` file has changed?  Sounds like a good idea :+1:

bogdan:
@steakchaser it should already be working like that. Do you have any issues with that?

steakchaser:
We've always had to run `tmp:cache:clear` even when `routes.rb` changes.  I'll test it out again tomorrow to confirm.

steakchaser:
Hi @bogdan, confirmed that any time we change `routes.rb` we have to run `tmp:cache:clear` prior to compiling assets, otherwise the changes do NOT get picked up.  Thanks for looking into this!

le0pard:
@steakchaser we need also rails version to check and reproduce this on our side :)

le0pard:
@steakchaser should be fixed by using sprockets-rails < 3

PR #170, opened 2015-12-25
Fix tests for sprockets 3 and depend_on for rails

le0pard:


bogdan:
What is the reason?

le0pard:
@bogdan current master branch is red on CI, because sprokets-rails >= 3 broken some test cases. Right now I didn't found documentation how to use `depend_on` on version 3 and this can be issue  for this ticket: https://github.com/railsware/js-routes/issues/168 

On this line https://github.com/railsware/js-routes/blob/master/lib/js_routes/engine.rb#L16 for sprockets-rails >= 3 `Rails.application.assets` is `nil`

PikachuEXE:
Maybe take a look at 
https://github.com/fnando/i18n-js/blob/be0691874574655ce57bc75574034cb256ae7111/lib/i18n/js/engine.rb#L23-L36

le0pard:
@PikachuEXE thanks - will check.

PR #172, opened 2015-12-26
fix named routes for rails 5

### LINKED ISSUE #171: Add Rails 5.0 compatibility
Hello, 

I'm migrating a project from Rails 4.2.5 to Rails 5.0.0.beta1 and I'm getting the following error

ActionView::Template::Error: private method `routes' called for #<ActionDispatch::Routing::RouteSet::NamedRouteCollection:0x007fde7a5e4498>

Complete backtrace : https://gist.github.com/dlupu/8d122a4d8ff64ab5cc4f

BR
Dorian

vonagam:
Fixes #171

This solution is compatible with all rails versions presented in Appraisal.

If you need to use it with sprockets v3, you can use:
> gem 'js-routes', github: 'vonagam/js-routes', branch: 'rails5'

### DIFF
--- a/lib/js_routes.rb
+++ b/lib/js_routes.rb
@@ -59,7 +59,7 @@ class JsRoutes
 
     def generate(opts = {})
       # Ensure routes are loaded. If they're not, load them.
-      if Rails.application.routes.named_routes.routes.keys.empty?
+      if Rails.application.routes.named_routes.to_a.empty?
         Rails.application.reload_routes!
       end
 
@@ -133,7 +133,7 @@ class JsRoutes
   protected
 
   def js_routes
-    js_routes = Rails.application.routes.named_routes.routes.sort_by(&:to_s).flat_map do |_, route|
+    js_routes = Rails.application.routes.named_routes.to_a.sort_by(&:to_s).flat_map do |_, route|
       rails_engine_app = get_app_from_route(route)
       if rails_engine_app.respond_to?(:superclass) && rails_engine_app.superclass == Rails::Engine && !route.path.anchored
         rails_engine_app.routes.named_routes.map do |_, engine_route|


le0pard:
I am not sure, what this fix sprockets v3. How it fix `depend_on` problem in rails engine?

vonagam:
No, i did not fix specs. That's why i kept sprockets version at "< 3" and did not added rails edge back to Appraisals.

But i fixed issue with accessing named routes.

In my app, which uses rails 5 and sprockets 3, i use "JsRoutes.generate!" and those changes was enough to make it work.

le0pard:
@vonagam thanks, but this solution not work with Ruby < 2.1
build 324 - railsware js-routes - travis ci 2015-12-26 23-28-57

vonagam:
Sad.
Rails 5 have required ruby version at ">= 2.2.2".

le0pard:
@vonagam yep, I know. But this gem should support not only rails 5.

vonagam:
Yes, i know. That's why i said "Sad" )

vonagam:
You may create separate branch for rails 5 for the time being.

le0pard:
@vonagam yes, but it is not a best way. Better user backport feature way, if it not available. Example:

```
unless Enumerable.method_defined?(:to_h)
  
  def coerce_to_ary(obj)
    return obj if obj.kind_of?(Array)

    begin
      ret = obj.to_ary
    rescue Exception => e
      raise TypeError, "Coercion error: #{obj.inspect}.to_ary => Array failed:\n" \
                       "(#{e.message})"
    end
    raise TypeError, "Coercion error: obj.to_ary did NOT return a Array (was #{ret.class})" unless ret.kind_of? Array
    ret
  end

  module Enumerable
    def to_h(*args)
      h = {}
      each_entry(*args) do |key_value|
        key_value = coerce_to_ary(key_value)
        if key_value.size != 2
          raise ArgumentError, "element has wrong array length (expected 2, was #{key_value.size})"
        end
        h[ key_value[0] ] = key_value[1]
      end
      h
    end
  end
end
```

vonagam:
Or you can try dark techinques like:
> named_routes.instance_variable_get( "@routes" )
This may work in all ruby versions.

le0pard:
Nope, `instance_variable_get` only for debugging :)

vonagam:
Yes, you are right, cheap hacks are way to dark side )

le0pard:
@vonagam https://github.com/railsware/js-routes/pull/172#issuecomment-167364757 - try to use something like this to fix all tests.

vonagam:
Okay, backports seems to fix the issue with old ruby versions.

le0pard:
@vonagam yes, it is definitely not `add_development_dependency`, it needed on runtime of gem.

vonagam:
Yep, my bad. Fixed.

le0pard:
@vonagam thanks. @bogdan need review :)

bogdan:
I think we should not rely on backports gem. This gem only find for the "end consumers": code that doesn't serve as dependency to another code. Can we make it our own implementation?

le0pard:
@bogdan our implementation can looks like this: https://github.com/railsware/js-routes/pull/172#issuecomment-167364757

bogdan:
Why it can't be just `Rails.application.routes.named_routes.to_a.to_h`?

le0pard:
@bogdan `to_h` method available only from Ruby 2.1, but js-routes working also with ruby 1.9.3 and 2.0 :)

bogdan:
What about using `to_a` all the time then?

le0pard:
@bogdan Maybe, I don't see big problems. @vonagam just should change PR in this case

vonagam:
Done

PR #176, opened 2016-01-04
Add support for sprockets-rails 3.0.0

### LINKED ISSUE #175: Please add support for sprockets-rails 3.0.0

rosenfeld:
Resolves #175

### DIFF
--- a/app/assets/javascripts/js-routes.js.erb
+++ b/app/assets/javascripts/js-routes.js.erb
@@ -1,3 +1,3 @@
-//= jsroutes_depend_on_routes
 <%# encoding: UTF-8 %>
+<%= depend_on "file://#{Rails.root.join 'config', 'routes.rb'}" %>
 <%= JsRoutes.assert_usable_configuration! && JsRoutes.generate %>
--- a/app/assets/javascripts/js-routes.js.erb
+++ b/app/assets/javascripts/js-routes.js.erb
@@ -1,3 +1,3 @@
 <%# encoding: UTF-8 %>
-<% depend_on "file://#{Rails.root.join 'config', 'routes.rb'}" %>
+<% depend_on "file://#{Rails.root.join 'config', 'routes.rb'}" if JsRoutes::SPROCKETS3 %>
 <%= JsRoutes.assert_usable_configuration! && JsRoutes.generate %>
--- a/app/assets/javascripts/js-routes.js.erb
+++ b/app/assets/javascripts/js-routes.js.erb
@@ -1,2 +1,3 @@
+//= jsroutes_depend_on_routes
 <%# encoding: UTF-8 %>
 <%= JsRoutes.assert_usable_configuration! && JsRoutes.generate %>
--- a/app/assets/javascripts/js-routes.js.erb
+++ b/app/assets/javascripts/js-routes.js.erb
@@ -1,3 +1,3 @@
 <%# encoding: UTF-8 %>
-<%= depend_on "file://#{Rails.root.join 'config', 'routes.rb'}" %>
+<% depend_on "file://#{Rails.root.join 'config', 'routes.rb'}" %>
 <%= JsRoutes.assert_usable_configuration! && JsRoutes.generate %>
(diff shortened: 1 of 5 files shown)


rosenfeld:
The failing example for Rails 3.2 wasn't introduced by this PR.

le0pard:
Thanks, @rosenfeld . BTW, @PikachuEXE propose to look at this way https://github.com/fnando/i18n-js/blob/be0691874574655ce57bc75574034cb256ae7111/lib/i18n/js/engine.rb#L23-L36 

P.S. I hope @bogdan will fix broken master.

rosenfeld:
Take a look at the comments for that file. It points to the `finisher_hook`. Take a look at its documentation:

http://guides.rubyonrails.org/configuring.html#initializers

"finisher_hook Provides a hook for after the initialization of process of the application is complete, as well as running all the config.after_initialize blocks for the application, railties and engines."

Since `after_initialize` is already enough for us since we only need to add another method to `Sprockets::DirectiveProcessor`, which doesn't depend on anything but ensuring Sprockets is loaded (which should be true by `after_initialize`), I think it should be fine. Sprockets doesn't provide any load hook that we could use in a regular initializer, that's why I changed to an `after_initialize` block.

rosenfeld:
Another alternative would be to explicitly require 'sprockets/directive_processor' but it would prevent any autoload on this class intended to speed up the boot process in case js-routes engine is loaded but sprockets is not for some reason... But since I didn't really check whether `::Sprockets` was defined, it would fail in this case. But anyway, I'm afraid of requiring sprockets before proper initialization...

rosenfeld:
Please take a look at my last commit for another approach which doesn't involve the need for any initializer. What do you think?

le0pard:
@rosenfeld looks so simple. Need to check what this really works if route file changed.

rosenfeld:
I've already checked but feel free to confirm :)

bogdan:
@rosenfeld which versions of rails did you check? I am fine if you are confident it works for all major versions since 3.2. Maybe makes sense to leave old method for sprockets < 3.0. Wdyt?

app/assets/javascripts/js-routes.js.erb
```diff
@@ -1,2 +1,3 @@
 <%# encoding: UTF-8 %>
+<%= depend_on "file://#{Rails.root.join 'config', 'routes.rb'}" %>
```
bogdan: What is the outcome of `depend_on`? What does it write the JS file?
I saw sprockets describes `depend_on` as a directive. What is the difference of using it like ruby method instead of `// depend_on` sprocket directive?

rosenfeld: Nothing, from my tests. I've based on this test of sprockets:

https://github.com/rails/sprockets/commit/763d2a956d61705b63a0aba3608392d38bdafeb1

This test was added to sprockets since 3.0.0.

rosenfeld:
I've added another commit to replace `<%=` with `<%` and another one to keep supporting sprockets < 3.

jaredbeck:
Thanks for supporting multiple versions of sprockets! It really helps avoid dependency conflicts.

Are there plans to cut a new release of js-routes any time soon? Thanks! For now, my project must stay on 1.2.0 until there's a release with sprockets 3 support.

bogdan:
Released 1.2.2. Enjoy.

rosenfeld:
Thanks!

PikachuEXE:
Would be great to update CHAGNELOG :)

js-routes.gemspec
```diff
@@ -21,7 +21,7 @@ Gem::Specification.new do |s|
   s.summary = %q{Brings Rails named routes to javascript}
 
   s.add_runtime_dependency(%q<railties>, [">= 3.2"])
-  s.add_runtime_dependency(%q<sprockets-rails>, ["< 3"])
+  s.add_runtime_dependency(%q<sprockets-rails>)
```
PikachuEXE: I think this is too open
Being `< 4` until 4.x is actually out is better

bogdan: I never believed in those limitations: my gems get broken on a random version update from a vendor - minor, major or patch. Especially js-routes - it went broken on minor version of rails several times. That's why I think those up level limitations never help.

PR #183, opened 2016-03-13
Sort by the route key instead of the entire AST.

### LINKED ISSUE #181: SystemStackError with gem
Hi there, I just set up the gem as given in the Readme, and when restarting my server, I get

```
<snip>
vendor/bundle/ruby/2.3.0/gems/activesupport-4.2.5.2/lib/active_support/core_ext/array/conversions.rb:98:in `inspect'
vendor/bundle/ruby/2.3.0/gems/activesupport-4.2.5.2/lib/active_support/core_ext/array/conversions.rb:98:in `to_formatted_s'
vendor/bundle/ruby/2.3.0/gems/js-routes-1.2.4/lib/js_routes.rb:153:in `each'
vendor/bundle/ruby/2.3.0/gems/js-routes-1.2.4/lib/js_routes.rb:153:in `sort_by'
vendor/bundle/ruby/2.3.0/gems/js-routes-1.2.4/lib/js_routes.rb:153:in `js_routes'
```

`Rails.application.routes.named_routes.to_a` returns an array of pairs, with each pair being the name of the route and the ActionDispatch::Journey::Route. Calling #to_s on any of those pairs seem to be triggering a StackError.

I'm using Rails 4.2.5.2. Is there anything I can do to work around this?

lowjoel:
This fixes #181.

Sorting `Rails.application.routes.named_routes.to_a` by `to_s` was introduced in d7dcfd45ae59dd8bea595f3bc5598567a0ec82f5, which was supposed to sort every route by key.

Rails.application.routes.named_routes.to_a.first.to_s returns a string like this:

```
"[:root, #<ActionDispatch::Journey::Route:0x0000a8ecc362f8 @name="root", @app=#<ActionDispatch::Routing::RouteSet::Dispatcher:0x0000a8ecc370b8 @defaults={:controller=>"application", :action=>"index"}>, @path=#<ActionDispatch::Journey::Path::Pattern:0x0000a8ecc36f78 @spec=#<ActionDispatch::Journey::Nodes::Slash:0x0000a8ecc486d8 @left="/", @memo=nil>, @requirements={}, @separators="/.?", @anchored=true, @names=[], @optional_names=[], @required_names=[], @re=nil, @offsets=nil>, @constraints={:required_defaults=>[:controller, :action], :request_method=>/^GET$/}, @defaults={:controller=>"application", :action=>"index"}, @required_defaults=nil, @required_parts=[], @parts=[], @decorated_ast=nil, @precedence=1, @path_formatter=#<ActionDispatch::Journey::Format:0x0000a8ecc36280 @parts=["/"], @children=[], @parameters=[]>>]"
```

This change will sort the routes using only their path, ignoring the ActionDispatch::Journey::Route object.

### DIFF
--- a/lib/js_routes.rb
+++ b/lib/js_routes.rb
@@ -150,7 +150,7 @@ class JsRoutes
   protected
 
   def js_routes
-    js_routes = Rails.application.routes.named_routes.to_a.sort_by(&:to_s).flat_map do |_, route|
+    js_routes = Rails.application.routes.named_routes.to_a.sort_by(&:first).flat_map do |_, route|
       rails_engine_app = get_app_from_route(route)
       if rails_engine_app.respond_to?(:superclass) && rails_engine_app.superclass == Rails::Engine && !route.path.anchored
         rails_engine_app.routes.named_routes.map do |_, engine_route|


le0pard:
@lowjoel let's add test case :)

lowjoel:
I'm not sure how this can be tested.

I can't get it to break since it requires a Rails route to be materialised and memoised within the Rails application, which is beyond the control of the gem unless you'd want to start digging into Rails internals?

bogdan:
@lowjoel are you sure that `first` is always a string? Are you sure we don't need `first.to_s`?

lowjoel:
@bogdan `first` is a Symbol, like routes.

And in Ruby's docs:

> to_s → string
> Returns the name or string corresponding to sym.

bogdan:
@lowjoel let's think over night about a test case and if no fresh idea - we will just merge.

lowjoel:
Yes, let me know if you guys have ideas too. Thanks @le0pard and @bogdan.

lowjoel:
@bogdan sorry, I can't think of anything.

bogdan:
I am on vacation now without my laptop. I'll merge n release next week

PR #185, opened 2016-03-31
changed export variable to request another object to avoid conflicts with other libraries

fedebonisconti:
Hi,

I have fixed an issue with [another library](https://secure.mlstatic.com/sdk/javascript/v1/mercadopago.js) that defines a _exports_ variable in global scope too. It can be useful for another external js libraries that expects to use the same variable at the global scope.

Cheers!


bogdan:
Sorry, I am a total noob as in backend JS and don't understand the purpose of that `exports` at all. Can you point me an article where it is described? @le0pard maybe you can help?

Your PR breaks the build, please fix it first.

le0pard:
Hello @fedebonisconti . As I know js-routes already using namespace as `Routes` - https://github.com/railsware/js-routes/pull/121#issuecomment-54075048 Why in this case it conflicts with another library? `require` commonjs and `import` from ES2015 is safe for this case.

le0pard:
@fedebonisconti ping

le0pard:
@fedebonisconti ping again

le0pard:
No response. Closing

PR #187, opened 2016-04-14
Default config.prefix to the Rails relative URL root and note details in the README

tknarr:


bogdan:
Is the ENV var reflected in rails setting `relative_url_path`? If so, I think we can make a default `prefix` option to qual to `relative_url_path` rails config option.

tknarr:
Yes, it is. From the discussions under the Rails issues it seems like it's only purpose is to get that setting correct during precompilation.

bogdan:
So, lets just do that.

tknarr:
OK, I'll add that to this pull request.

tknarr:
I'm just going to start a different pull request after I set things up correctly on my end.

Issue #186, opened 2016-04-14
prefix setting not being applied


tknarr:
This seems related to, or maybe the same thing as, issue #104. I use the initializer from the README, and set config.prefix to Rails.application.config.relative_url_root. In development it works fine, the prefix shows up in the paths generated by the `*_path()` functions as expected. In production, the prefix doesn't appear. I tested it with a hard-coded string and got identical results. I'm not using Heroku or anything, it's a vanilla setup: Apache and Passenger for the server, deploying via RubyMine's built-in SFTP tool (confirmed that the results are what I expected) and using the latest version of js-routes (1.2.5) with Ruby 2.2.4 and Rails 4.2.6. I made sure to stop Apache and run the `tmp:clear` and `assets:clobber` Rake tasks before doing the `assets:precompile` step, all with RAILS_ENV=production set. The initializers are running during the precompile, I can see the output and a `puts` statement in my initializer shows js-route's config.prefix set to the expected string "/CertificateAuthority" for the sub-URI I have the application installed in.

I'm going to dig down into the code to see if I can figure out where the prefix is going missing and why it only affects production and not development.

le0pard:
I think problem in caching of assets. If you change initializer values, it will not change content of js files and sha1 checksum for this files will be the same. That is why it will be not recompile.

tknarr:
I thought about that, that's why I did the `tmp:clear` and `assets:clobber` tasks to clear out all the precompiled assets and the cache before I did the precompilation. It's got to be something to do with precompilation, I've checked backwards and forwards and that's the only difference between the two environments I can find, but I don't know exactly what. Yet. With luck it'll turn out to be something I didn't realize I fumbled.

le0pard:
You can also try to use erb with ruby generator:
```
# app/assets/javascripts/routes.js.erb
<%= JsRoutes.generate(prefix: "/CertificateAuthority") %>
```

tknarr:
[...]

PR #189, opened 2016-04-15
Default prefix setting to Rails relative URL root setting

tknarr:
Issue #186:
Take the default for the `prefix` configuration setting from the Rails relative URL root setting (`Rails.application.config.relative_url_root`).
Add a note in the README about how to set the relative URL root correctly during asset precompilation.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -58,7 +58,7 @@ Available options:
   * Default: `Routes`
 * `prefix` - String representing a url path to prepend to all paths.
   * Example: `http://yourdomain.com`. This will cause route helpers to generate full path only.
-  * Default: value of Rails.application.config.relative_url_root or RAILS_RELATIVE_URL_ROOT env variable
+  * Default: Rails.application.config.relative_url_root
 * `camel_case` (version >= 0.8.8) - Generate camel case route names.
   * Default: false
 * `url_links` (version >= 0.8.9) - Generate `*_url` helpers (in addition to the default `*_path` helpers).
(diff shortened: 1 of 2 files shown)


bogdan:
Can we write a test for that?

Readme.md
```diff
@@ -58,7 +58,7 @@ Available options:
   * Default: `Routes`
 * `prefix` - String representing a url path to prepend to all paths.
   * Example: `http://yourdomain.com`. This will cause route helpers to generate full path only.
-  * Default: blank
+  * Default: value of Rails.application.config.relative_url_root or RAILS_RELATIVE_URL_ROOT env variable
```
bogdan: Default: `Rails.application.config.relative_url_root`. I think this is enough. People that use ENV aware of that.

Readme.md
```diff
@@ -73,6 +73,12 @@ Available options:
   * Example: `jQuery.param` - use jQuery's serializer algorithm. You can attach serialize function from your favorite AJAX framework.
   * Example: `MyApp.custom_serialize` - use completely custom serializer of your application.
 
+**Important:** If you've installed your application in a sub-path or sub-URI of your server instead of at the root, you need to set the `RAILS_RELATIVE_URL_ROOT` environment variable to the correct path prefix for your application when you precompile assets. Eg., if your application's base URL is "https://appl.example.com/Application1", the command to precompile assets would be:
+```
+RAILS_RELATIVE_URL_ROOT=/Application1 RAILS_ENV=production bundle exec rake assets:precompile
+```
+The environment variable is only needed for precompilation of assets, at any other time (eg. when assets are compiled on-the-fly as in the development environment) Rails will set the relative URL root correctly on it's own.
+
```
bogdan: I think it should go after `Very Advanced Setup` section. `Important` label makes it feel like it is improtant to everyone. But I think this is only important to 1% of rails users.

Let's just make a regular paragraph like `### Rails Relative Url Root`

tknarr:
I looked at how to test it, but I don't know of any way to get Rails' test system to run the application at anywhere but the document root and without that the test always passes even if the code's broken. I can rummage around to find out where to pull the SCRIPT_NAME out to compare the prefix setting to it instead of a hard-coded check, I can't adjust the Travis configuration to get a test run that'll exercise that code. :(

bogdan:
Can you change the rails config value, generate JS, make an assert and then put back old config value?

tknarr:
Unfortunately not. The `prefix` setting gets set during Rails initialization, any changes to `relative_url_path` after that don't affect js-routes. I can't set `relative_url_path` before Rails initialization either, whatever I set just gets overwritten based on the env variable or lack of same. I have to manipulate the environment to get the variable set before Rails initialization. I think I've figured out how to do it, and learned enough Rspec to write the test correctly.

bogdan:
Ok, let's leave the test. Just small fixes in readme and we are good to go.

tknarr:
Done.

PR #193, opened 2016-06-02
Drop 1.9.3 support and add indexOf helper

le0pard:
Reasons:
 
* New v8 engine (5.x instead 3.x in therubyracer gem)
* Simple interface, which provide better way to debug JS output (result - `log` function much simpler)

Reason to remove "no `indexOf`" test:

* no engines, which is not have this basic function - https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf look at "Browser compatibility" section
* we already using `indexOf` in another places and not checking availability. Example: https://github.com/railsware/js-routes/blob/master/lib/routes.js.coffee#L112

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,10 @@
 language: ruby
 cache: bundler
 
+before_install:
+  - gem install bundler # need for jruby and ruby-head
+
 rvm:
-  - 1.9.3
   - 2.0
   - 2.1
   - 2.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## master
 
+* Remove support 1.9.3
+* Add helper for indexOf, if no native implementation in JS engine
+
 ## v1.2.6
 
 * Use default prefix from `Rails.application.config.relative_url_root` #186
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## master
 
-* Remove support 1.9.3
+* Drop support 1.9.3
 * Add helper for indexOf, if no native implementation in JS engine
 
 ## v1.2.6
--- a/js-routes.gemspec
+++ b/js-routes.gemspec
@@ -29,9 +29,7 @@ Gem::Specification.new do |s|
   if defined?(JRUBY_VERSION)
     s.add_development_dependency(%q<therubyrhino>, [">= 2.0.4"])
   else
-    if RUBY_VERSION >= "2.0.0"
-      s.add_development_dependency(%q<byebug>)
-    end
+    s.add_development_dependency(%q<byebug>)
     s.add_development_dependency(%q<therubyracer>, [">= 0.12.1"])
   end
 end
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -173,7 +173,7 @@ Based on Rails routes of APP_CLASS
       for (key in options) {
         if (!hasProp.call(options, key)) continue;
         value = options[key];
-        if (ReservedOptions.indexOf(key) >= 0) {
+        if (this.indexOf(ReservedOptions, key) >= 0) {
           result[key] = value;
         } else {
           url_parameters[key] = value;
@@ -245,7 +245,7 @@ Based on Rails routes of APP_CLASS
       }
     },
     is_optional_node: function(node) {
-      return node === NodeTypes.STAR || node === NodeTypes.SYMBOL || node === NodeTypes.CAT;
+      return this.indexOf([NodeTypes.STAR, NodeTypes.SYMBOL, NodeTypes.CAT], node) >= 0;
     },
     build_path_spec: function(route, wildcard) {
       var left, right, type;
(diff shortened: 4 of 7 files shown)


bogdan:
I am against relaying on `indexOf`. The current usage of it should be removed. This is recently added thing and huge apps maintainers will complain after updating js-routes. As far as I remember we also support IE8 in some of our projecs.

I also don't see a lot of reasons on moving from to `mini_racer`.  It looks like a change for a change. It is less popular gem that may mean that it is more buggy.

le0pard:
BTW, @bogdan, maybe we should drop support Ruby 1.9.3 (support end at February 23 2015)

bogdan:
1.9.3 removal is ok.

le0pard:
@bogdan maybe I can add polyfill for `indexOf` and in this case it will used, if no native support of it. What do you think?

bogdan:
what is polyfill?

le0pard:
A polyfill, or polyfiller, is a piece of code (or plugin) that provides the technology that you, the developer, expect the browser to provide natively. Example:

```
Base64 = (->

  # Character map
  charMap = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/="

  # Encode
  encode = @btoa || (input) ->

    output = ""
    i      = 0

    while i < input.length

      chr1 = input.charCodeAt i++
      chr2 = input.charCodeAt i++
      chr3 = input.charCodeAt i++

      enc1 = chr1 >> 2
      enc2 = ((chr1 & 3) << 4) | (chr2 >> 4)
      enc3 = ((chr2 & 15) << 2) | (chr3 >> 6)
      enc4 = chr3 & 63

      if isNaN chr2 then enc3 = enc4 = 64
      else if isNaN chr3 then enc4 = 64

      for char in [ enc1, enc2, enc3, enc4 ]
        output += charMap.charAt char

    output

  # Decode
  decode = @atob || (input) ->

    # remove invalid chars
    input  = input.replace /[^A-Za-z0-9\+\/\=]/g, ""
    i      = 0
    output = ""

    while i < input.length

      enc1 = charMap.indexOf input.charAt i++
      enc2 = charMap.indexOf input.charAt i++
      enc3 = charMap.indexOf input.charAt i++
      enc4 = charMap.indexOf input.charAt i++

      chr1 = (enc1 << 2) | (enc2 >> 4)
      chr2 = ((enc2 & 15) << 4) | (enc3 >> 2)
      chr3 = ((enc3 & 3) << 6) | enc4

      output += String.fromCharCode chr1

      if enc3 != 64 then output += String.fromCharCode chr2
      if enc4 != 64 then output += String.fromCharCode chr3

    output

  ###*
  * Encodes the specified input to Base64.
  * @param {String} input
  * @return {String} result
  ###
  encode: (input) -> encode unescape( encodeURIComponent input )

  ###*
  * Decodes the specified input from Base64.
  * @param {String} input
  * @return {String} result
  ###
  decode: (input) -> decodeURIComponent escape( decode input )

)()
```

Class trying to use `atob` and `btoa` for base64, but if no - have own implementation.

bogdan:
Does it mean monkeypatch Array?

le0pard:
yes, add `indexOf` of not present. Or I can create helper, which will use `indexOf` or non native implementation. Which is better?

bogdan:
`indexOf: (array, element) ->` is just fine. It isn't worth it.... leave it as is....

le0pard:
Added helper for `indexOf` and back tests.

Additional points for mini_racer (but if you still don't want - I will remove it):

* therubyracer currently (May 2016) only supports v8 version 3.16.14
* Creator of mini_racer - Sam Saffron (discourse, MiniProfiler, fast_blank)
* It was created as an alternative to the excellent therubyracer. Unlike therubyracer, mini_racer only implements a minimal bridge. This reduces the surface area making upgrading v8 much simpler and exhaustive testing simpler (less code - less bugs)
* Speed:
```
$ ruby bench_uglify.rb
Benching with MiniRacer
MiniRacer minify discourse_app.js 13813.36ms
MiniRacer minify discourse_app_minified.js 18271.19ms
MiniRacer minify discourse_app.js twice (2 threads) 13587.21ms

Benching with therubyracer
MiniRacer minify discourse_app.js 151467.164ms
MiniRacer minify discourse_app_minified.js 158172.097ms
MiniRacer minify discourse_app.js twice (2 threads) - DOES NOT FINISH

Killed: 9
```

As you can see in result `log` function for debugging JS right now just `proc{|value| puts value.inspect }` in ruby :)

bogdan:
> * therubyracer currently (May 2016) only supports v8 version 3.16.14

I think version 1.x already had enough functionality for js-routes

> * Creator of mini_racer - Sam Saffron (discourse, MiniProfiler, fast_blank)

So what?

> * It was created as an alternative to the excellent therubyracer. Unlike therubyracer, mini_racer only implements a minimal bridge. This reduces the surface area making upgrading v8 much simpler and exhaustive testing simpler (less code - less bugs)

I don't follow you here. We never had any problems upgrading therubyracer... we didn't even have a necessity for it.

> * Speed:

We are not concerned about speed in JS routes at all

> As you can see in result `log` function for debugging JS right now just `proc{|value| puts value.inspect }` in ruby :)

This is the only true benefit I see.

le0pard:
Ok. So what is your final decision? Should mini_racer 👍  or 👎 ?

bogdan:
👎  not enough benefits.

le0pard:
@bogdan done :)

PR #196, opened 2016-07-01
fix Sprockets3 compatibility

sevaorlov:
It fixed the problem for our app, but will be good if someone else can try it one more time.

Continuation of https://github.com/railsware/js-routes/pull/170

### DIFF
--- a/lib/js_routes/engine.rb
+++ b/lib/js_routes/engine.rb
@@ -4,15 +4,27 @@ class JsRoutes
   SPROCKETS3 = Gem::Version.new(Sprockets::VERSION) >= Gem::Version.new('3.0.0')
   class Engine < ::Rails::Engine
 
-    initializer 'js-routes.dependent_on_routes', after: "sprockets.environment" do
-
-      if Rails.application.assets.respond_to?(:register_preprocessor)
-        routes = Rails.root.join('config', 'routes.rb').to_s
-        Rails.application.assets.register_preprocessor 'application/javascript', :'js-routes_dependent_on_routes' do |ctx,data|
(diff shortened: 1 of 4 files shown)


lib/js_routes/engine.rb
```diff
@@ -4,15 +4,27 @@ class JsRoutes
   SPROCKETS3 = Gem::Version.new(Sprockets::VERSION) >= Gem::Version.new('3.0.0')
   class Engine < ::Rails::Engine
 
-    initializer 'js-routes.dependent_on_routes', after: "sprockets.environment" do
+    if SPROCKETS3
+      initializer 'js-routes.dependent_on_routes', after: :engines_blank_point, before: :finisher_hook do
```
sevaorlov: https://github.com/fnando/i18n-js/pull/371/files#r45864633

le0pard:
@sevaorlov looks like tests is broken.

sevaorlov:
@le0pard, yep, I've added a commit to fix them

lib/js_routes/engine.rb
bogdan: Why do you combo `after` and `before` options at the same time? What does it mean?

bogdan: @sevaorlov any response here?

sevaorlov: @bogdan, we need `after: :engines_blank_point` in order to run this block after engines are initialized, `finisher_hook` is used as border as before this hook all initializations should be completed http://guides.rubyonrails.org/configuring.html#initializers

PR #197, opened 2016-07-11
Don't apply defaults to path

krzysiek1507:
We don't want to apply defaults to path because they are backend only

### DIFF
--- a/lib/js_routes.rb
+++ b/lib/js_routes.rb
@@ -36,6 +36,7 @@ class JsRoutes
 
   LAST_OPTIONS_KEY = "options".freeze
   FILTERED_DEFAULT_PARTS = [:controller, :action, :subdomain]
+  URL_OPTIONS = [:protocol, :domain, :host, :port, :subdomain]
 
   class Options < Struct.new(*DEFAULTS.keys)
     def to_hash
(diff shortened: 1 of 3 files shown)


lib/js_routes.rb
```diff
@@ -36,6 +36,7 @@ class JsRoutes
 
   LAST_OPTIONS_KEY = "options".freeze
   FILTERED_DEFAULT_PARTS = [:controller, :action, :subdomain]
+  URL_OPTIONS = [:protocol, :domain, :host, :port]
```
bogdan: Why isn't `subdomain` in this hash? Why is it processed differently from `domain`?

krzysiek1507: It could be there, but `subdomain` is also in `FILTERED_DEFAULT_PARTS`, so the check will be 
```ruby
FILTERED_DEFAULT_PARTS.exclude?(:subdomain) && URL_OPTIONS.include?(:subdomain)
```
and this is always false. Of course, I can add it.

bogdan: If you want to add it, please add a test that would have failed without it.

krzysiek1507: Do you have any failing test?

bogdan: No, I don't mean that. Please add a test that test for inclusion of subdomain in `URL_OPTIONS` if it exists.

krzysiek1507: None of the tests is failing when subdomain exists in `URL_OPTIONS`.
I've added a subdomain to `URL_OPTIONS`.

krzysiek1507:
Thanks! Are you going to release new version?

bogdan:
Sure, want to also have 5.0 support in place from a different PR

PR #202, opened 2016-07-28
Update Sprockets extension to work with Sprockets 2, 3, 4

flood4life:
Sprockets 3.x display deprecation warning: You are using the a deprecated processor interface #<Proc:0x007f9002044f90@/.../gems/ruby-2.3.1/gems/js-routes-1.2.7/lib/js_routes/engine.rb:11>.
Please update your processor interface:
https://github.com/rails/sprockets/blob/master/guides/extending_sprockets.md#supporting-all-versions-of-sprockets-in-processors

I've followed the guide and updated the engine to support all versions of sprockets.

This makes two tests fail, but it looks like it's because those tests are implementation specific. I've checked this code on three fresh applications with sprockets 2.12.4, 3.7.0 and 4.0.0beta2. JavaScript `Routes` object is correctly created in all three cases.

### DIFF
--- a/lib/js_routes/engine.rb
+++ b/lib/js_routes/engine.rb
@@ -32,7 +32,7 @@ class Engine < ::Rails::Engine
   v2                = Gem::Dependency.new('', ' ~> 2')
   v3                = Gem::Dependency.new('', ' ~> 3')
   v4                = Gem::Dependency.new('', ' >= 4')
-  sprockets_version = Sprockets::VERSION
+  sprockets_version = Gem::Version.new(Sprockets::VERSION).release
   initializer_args  = case sprockets_version
                         when -> (v) { v2.match?('', v) }
                           { after: "sprockets.environment" }
--- a/lib/js_routes/engine.rb
+++ b/lib/js_routes/engine.rb
@@ -29,19 +29,41 @@ end
 
 class Engine < ::Rails::Engine
   require 'sprockets/version'
-  sprockets3       = Gem::Version.new(Sprockets::VERSION) >= Gem::Version.new('3.0.0')
-  initializer_args = if sprockets3
-                       { after: :engines_blank_point, before: :finisher_hook }
-                     else
-                       { after: "sprockets.environment" }
-                     end
+  v2                = Gem::Dependency.new('', ' ~> 2')
+  v3                = Gem::Dependency.new('', ' ~> 3')
+  v4                = Gem::Dependency.new('', ' >= 4')
+  sprockets_version = Sprockets::VERSION
+  initializer_args  = case sprockets_version
+                        when -> (v) { v2.match?('', v) }
+                          { after: "sprockets.environment" }
+                        when -> (v) { v3.match?('', v) || v4.match?('', v) }
+                          { after: :engines_blank_point, before: :finisher_hook }
+                        else
+                          raise StandardError, "Sprockets version #{sprockets_version} is not supported"
+                      end
 
   initializer 'js-routes.dependent_on_routes', initializer_args do
-    env = if sprockets3
-            Sprockets
-          else
-            Rails.application.assets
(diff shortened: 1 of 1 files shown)


bogdan:
You need to ensure that it not only words but also updates automatically when you change the routes.rb file.

I am not sure I want to change anything for older version of Sprockets. There were so many concerns before and let's leave them as they were before.

flood4life:
The point is not the older versions of Sprockets, it's the Sprockets 4, which will remove the currently used block interface.

Sure, I'll check later that it updates on `routes.rb` change.

bogdan:
I know that you are targeting Sprockets 4, but I don't want to change behaviour for Sprockets < 4 because it is verified that it is working there and I don't want to fix bugs for 3 and 2 after this patch.

flood4life:
There, I've verified that JS routes are updated on `routes.rb` change. All tests are passing, and I've double checked this in Rails 3.2 (Sprockets 2) and Rails 4-5 (Sprockets 2, 3, 4).

Well, sooner or later you'll have to update this to support Sprockets 4, so why not now?

bogdan:
Sure, I want to update for 4 now.... but I don't want to change anything for 2 and 3. Can you leave the code as is for those with top level case by version statement?

flood4life:
Well, that wouldn't be DRY, and there are actually no changes for 2: it uses the same old `new.render` interface. I even found a mistake in Sprockets upgrade guide (and proposed a fix) while ensuring that it works the same.

If you're not convinced, though, I can add the case statement next week, I think.

bogdan:
There were so many caveats with 2 and 3. I want us to look forward and maintain the functionality for 4 without thinking if it supports 2 and 3 that is why I want to 2 and 3 to be frozen forever. No DRY is OK in this case as we will not maintain 2 and 3 at all.

flood4life:
Done. Sprockets 2-3 now uses the old code, and sprockets >= 4 uses new interface. However, I was not certain with condition for sprockets > 4. What do you think, is '>= 4' condition ok for now? Or should we also enforce '< 6'?

bogdan:
I like to be optimistic in version support... so `>= 4` sounds good to me.

lib/js_routes/engine.rb
```diff
@@ -1,30 +1,69 @@
...
+                          { after: "sprockets.environment" }
+                        when -> (v) { v3.match?('', v) || v4.match?('', v) }
+                          { after: :engines_blank_point, before: :finisher_hook }
+                        else
+                          raise StandardError, "Sprockets version #{sprockets_version} is not supported"
+                      end
+
+  initializer 'js-routes.dependent_on_routes', initializer_args do
+    case sprockets_version
+      when -> (v) { v2.match?('', v) }
```
bogdan: Are you sure this syntax is supported by all versions of ruby that we do support?

flood4life: Are you referring to lambdas in `case`? It is available since Ruby 1.9.

bogdan: Yes, thanks.

lib/js_routes/engine.rb
```diff
@@ -1,30 +1,69 @@
...
+  v2                = Gem::Dependency.new('', ' ~> 2')
+  v3                = Gem::Dependency.new('', ' ~> 3')
+  v4                = Gem::Dependency.new('', ' >= 4')
+  sprockets_version = Gem::Version.new(Sprockets::VERSION).release
+  initializer_args  = case sprockets_version
+                        when -> (v) { v2.match?('', v) }
+                          { after: "sprockets.environment" }
+                        when -> (v) { v3.match?('', v) || v4.match?('', v) }
+                          { after: :engines_blank_point, before: :finisher_hook }
+                        else
```
mamantoha: What about Sprockets 4?

PR #214, opened 2016-11-01
Added `application` option

ankane:
Hey, first off, thanks for the awesome work on this gem! 👍 

This adds the ability to generate routes from a Rails engine.

```erb
<%= JsRoutes.generate(engine: Blazer::Engine) %>
```

I've attempted to use this for [Blazer](https://github.com/ankane/blazer), but am running into issues with different route names depending on where users mount the engine. If mounted at `/`, it generates `Routes.blazer_queries_path()`, but if namespaced under `admin`, it generates `Routes.admin_blazer_queries_path()`, which makes sense for the typical setup.

Let me know your thoughts. Thanks!

### DIFF
--- a/lib/js_routes.rb
+++ b/lib/js_routes.rb
@@ -103,7 +103,7 @@ class JsRoutes
   def generate
     {
       "GEM_VERSION"         => JsRoutes::VERSION,
-      "APP_CLASS"           => Rails.application.class.to_s,
+      "APP_CLASS"           => application.class.to_s,
       "NAMESPACE"           => @options[:namespace],
       "DEFAULT_URL_OPTIONS" => json(@options[:default_url_options].merge(deprecate_url_options)),
       "PREFIX"              => @options[:prefix] || Rails.application.config.relative_url_root || "",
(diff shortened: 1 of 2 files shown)


bogdan:
The idea looks good to me. 
Some changes required. 

* Why do you think we should pass an engine instead of an app? I would prefer to use an App only.
* There are some more `Rails.application` usages. Why didn't you replace them with new `app` method call.
* Please add a test.

ankane:
Hey @bogdan, thanks for the super quick response.

1. I think `:engine` would be DRYer, since I'm not sure you'd pass another app that's not an engine, but happy to make it `:app` instead. (no strong feelings either way)
2. Wasn't sure if it's a good idea to update the other instances. One ensures routes are loaded (probably safe to update that one) and the other ensures the app is set to initialize on precompile (think we need to keep this as is).
3. Will do

bogdan:
1. `application` feels the best option name for me because `Rails.application` as a default value looks pretty solid and `engine` feel more specific. You just generate routes for the given application. That feels logical instead of you may generate routes for the current rails application or specify a custom engine where you would generate a routes. 
2. Why do we need to keep `initialize_on_precompile` from the Rails application ? Do you think that any custom engine can relay on that?

ankane:
Hey @bogdan,

1. After testing, it looks like it's better to pass the engine directly as the application rather than `MyEngine.app`, so `application` is a better name.

2. Not sure what you mean by "relay" that. An engine should always be mounted inside a Rails application and won't respond to `config.relative_url_root` or `reload_routes!`, so I believe we want to use `Rails.application` in those places.

bogdan:
2. Can a rails engine be run standalone without parent rails application? I thought it is possible.
3. Please add a doc to readme.

bogdan:
I reworked the code slightly and updated readme. Thanks.

ankane:
Great, thanks @bogdan 👍

PR #216, opened 2017-01-17
Improve optional parameters support.

marshall-lee:


lib/js_routes.rb
```diff
@@ -216,13 +216,19 @@ def build_js(route, parent_route)
...
       FILTERED_DEFAULT_PARTS.exclude?(part) && URL_OPTIONS.include?(part) || required_parts.include?(part)
     end
     [
-      required_parts, optional_parts, serialize(route.path.spec, parent_spec), default_parts
+      parts,
+      required_parts,
+      optional_parts,
+      default_options,
+      Rails.version < '4.0',
+      serialize(route.path.spec, parent_spec)
```
bogdan: Js-Routes is optimized to generate as small amount of JS as possible per rails route. That is why number of route arguments needs to be minified:

1. Pass `Rails.version < '4.0'` through configuration but a constant argument of each route.
2. You don't need to pass `parts  required_parts and optional_parts` at the same time. Leave it as it was with only required and optional parts and concat them in JS to receive `parts`.

marshall-lee: @bogdan Hi, thanks for your feedback!

I don't agree about keeping only `required_parts` and `optional_parts` as it was before, third parameter was introduced for a reason.
It's impossible to merge these two arrays into original `parts` because the order of elements is unknown. Suppose something like `/(:opt1)/:arg1/(:opt2)/:arg2` — how do I know that `:arg1` comes before `:opt2` having only `[:opt1, :opt2]` and `[:arg1, :arg2]`? The order of elements in `parts` is crucial for correctness.

Instead, I propose to pass only `parts` and `required_parts` so I can restore `optional_parts` from it. Passing `parts` and `optional_parts` is also an option (in this case `required_parts` will be restored) but I assume that `required_parts` is usually shorter than `optional_parts` so I suggest to optimize with this assumption in the head.

marshall-lee: There's also an option to pass a table like `{"opt1":0,"arg1":1,"opt2":0,"arg2":1}` (I used `0` in `1` because they're shorter than boolean literals) but as I know JavaScript objects don't maintain the order of fields which is crucial here.

bogdan: I like the hash idea. We can use `{"opt1": false,"arg1":true,"opt2":false,"arg2":true}` because `false` and `true` turn into `!0` and `!1` using JS minification. I think this is good enough.

marshall-lee: @bogdan But JavaScript objects don't guarantee the order of fields as I said... They're like hashes in Ruby < 1.9. Only arrays preserve the order.

marshall-lee: Definition of an Object from [ECMAScript Third Edition (pdf)](http://www.ecma-international.org/publications/files/ECMA-ST-ARCH/ECMA-262,%203rd%20edition,%20December%201999.pdf):

> 4.3.3 Object

> An object is a member of the type Object. It is **an unordered collection of properties** each of which contains a primitive value, object, or function. A function stored in a property of an object is called a method.

So the order of properties in JSON serialization could differ from the order of properties in deserialized object when iterating it using `for`. Order is very crucial as I said.

marshall-lee: The only option is an array `[['opt1', false], ['arg1', true], ['opt2', false], ['arg2', true]]`.

bogdan: Yes, that's fine. Lets do that.

marshall-lee:
@bogdan I made all necessary changes and rebased the branch.

spec/support/routes.rb
```diff
@@ -26,7 +26,9 @@ def draw_routes
     resources :returns
 
     scope "(/optional/:optional_id)" do
-      resources :things
+      resources :things do
+        get "deep/:deep_id(/:optional_deep_id)", to: "things#index", as: :deep
```
bogdan: I don't like parameters name here. The tests are harder to read because of it. Let's make a new route for your advanced tests like:

``` ruby
get (/:first_optional)/pages/:second_required/:third_required/edit(/:fourth_optional)
```

Better proposals welcome. I just want to make parameters order more logical so that you don't need to consult with route to understand the test.

PR #225, opened 2017-09-01
Fix JS checks

le0pard:


bogdan:
We need a test.

lib/routes.js.coffee
```diff
@@ -275,30 +275,27 @@ Utils =
...
 
   route_url: (route_defaults) ->
-    return route_defaults if typeof route_defaults == 'string'
+    return route_defaults if typeof route_defaults is 'string'
     protocol = route_defaults.protocol || Utils.current_protocol()
-    hostname = route_defaults.host || window.location.hostname
+    hostname = route_defaults.host || Utils.current_host()
     port = route_defaults.port || (Utils.current_port() unless route_defaults.host)
     port = if port then ":#{port}" else ''
+    "#{protocol}://#{hostname}#{port}"
```
krzysiek1507: it can produce:
`https://null:3000`

le0pard: @krzysiek1507 so if no hosts - no `"#{protocol}://#{hostname}#{port}"` at all? Or can be better solution?

krzysiek1507: @le0pard I was thinking about it and it seems that `hostname` has to be required for `_url` helpers in a `windowless` environment. It doesn't make sense to render HTML with URL like `https://null/products/bag` or `https:///products/bag`. Of course, it makes sense to render path but we want URL. Maybe we should follow Rails' implementation? What do you think?

krzysiek1507:
Hi @le0pard I checked it again and I think we should fall back to path when `hostname` is missing

krzysiek1507:
I've created PR #226

le0pard:
@krzysiek1507 looks good. Closing this one

PR #227, opened 2017-09-04
Cleanup Appraisals && test against Rails 5.1

krzysiek1507:
1. Sprockets 2.x can't be installed with Rails > 5
2. Test against Rails 5.1

Maybe we should stop testing against `ruby-head` and `jruby-head`? It takes a lot of time but I'm not sure if it is meaningful.


.gitignore
```diff
@@ -58,3 +58,4 @@ gemfiles/*.lock
 /spec/dummy/app/assets/javascripts/routes.js
 /spec/dummy/logs
 /spec/dummy/tmp
+.idea
```
bogdan: you should use local gitignore instead of addition your things to the repo git ignore. It can go forever otherwise....

Appraisals
```diff
@@ -1,16 +1,34 @@
...
+  sprockets2 rails, version
+  sprockets3 rails, version
+end
+
+{
+  rails50: '5.0.5',
+  rails51: '5.1.3'
+}.each do |rails, version|
+  sprockets3 rails, version
+end
```
bogdan: I would do something like this here:

``` ruby
[
  ['rails40', '4.0.13', [2,3]], 
  ['rails50, '5.0.5', [3]] ....
].each do |name, version, sprockets|
  define_appraisal(name, version, sprockets)
end
```

krzysiek1507:
@bogdan fixed

PR #235, opened 2018-01-11
Change ParameterMissing custom error implementation

wordofchristian:
With the current implementation of the ParameterMissing custom error type, function names can become obfuscated or dropped in the resulting stack trace. This PR updates the implementation based on the recommendations [from the mozilla developer docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#ES5_Custom_Error_Object).

## Example

I was able to reproduce the problem in our application with the help of [error-stack-parser](https://github.com/stacktracejs/error-stack-parser) to print the function names from the stacktrace.

```javascript
import Routes from 'routes';
import ErrorStackParser from 'error-stack-parser';

function doSomething() {
  // should take an argument
  Routes.account_path();
}

class SomeView {
  handleClick() {
    try {
      doSomething();
    } catch (error) {
      console.log(ErrorStackParser.parse(error).map(frame => frame.functionName));
    }
  }
}
```
screenshot 2018-01-10 16 00 34

Notice the function `doSomething` is not present in the stacktrace, nor is `handleClick`. Also this stacktrace contains a lot of webpack require code because it was generated at import/definition time rather than runtime.

Now with the change in this PR:

screenshot 2018-01-10 16 38 20

Notice that we can now see which route function caused the problem (`account_path`), as well as the calling function chain.

### DIFF
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -12,7 +12,11 @@ Based on Rails routes of APP_CLASS
 
   function ParameterMissing(message, fileName, lineNumber) {
     var instance = new Error(message, fileName, lineNumber);
-    Object.setPrototypeOf(instance, Object.getPrototypeOf(this));
+    if(Object.setPrototypeOf) {
+      Object.setPrototypeOf(instance, Object.getPrototypeOf(this));
+    } else {
+      instance.__proto__ = this.__proto__;
+    }
     if (Error.captureStackTrace) {
       Error.captureStackTrace(instance, ParameterMissing);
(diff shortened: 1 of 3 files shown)


bogdan:
Excellent! Do you think we can make a test for backtrace to include to route name by copy-pasting some essentials from `ErrorStackParser.parse` into the `js-routes` test suite?

wordofchristian:
done

PR #238, opened 2018-06-29
Avoid rewrite js file if content wasn't changed.

aleksejleonov:
Generating `routes.js` file using rake task rewrites it completely each time. So, asset pipeline or webpacker would recompile assets in case when content wasn't changed (because mtime of `routes.js` was changed).

This PR just skip write to output file, if file already exist and have same content. This way mtime became unchanged and asset pipeline or webpacker properly skip assets recompile.

Mostly same as https://github.com/fnando/i18n-js/pull/473 that solved same problem for `i18n-js` gem.

### DIFF
--- a/lib/js_routes.rb
+++ b/lib/js_routes.rb
@@ -150,8 +150,15 @@ class JsRoutes
     # https://github.com/railsware/js-routes/issues/7
     Rails.configuration.after_initialize do
       file_name ||= self.class.configuration['file']
-      File.open(Rails.root.join(file_name), 'w') do |f|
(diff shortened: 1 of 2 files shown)


le0pard:
Looks good. Waiting for @bogdan

aleksejleonov:
Read entire file is anyway better then rewrite entire file each time :)

Usually, routes file isn't very large. And adding hash to file would also add much complication to code.
Right now, I prefer to keep it simple. If somebody will experience problems with reading very large routes file — hash could be added later.

Issue #215, opened 2016-12-11
Switch from CoffeeScript to modular ES2015


valscion:
Hi there!

First of all, huge thanks for this gem, it has proven to be very useful to us!

I would really love this gem to work also for more complex and modern JavaScript use cases than simply using Sprockets to concatenate all JS source files together to a single blob.

Let me tell you about

1. How we use this gem
2. What issues the current implementation causes us
3. What is my proposal on how to do things differently

Rest assured that I'm not asking **you** to do these changes that I am proposing. I'm mainly interested in hearing your thoughts on whether you'd be willing to accept a big overhaul for the JS code of this gem as a future pull request, if our team could make this a reality.

I'd also be very interested to hear whether anyone else using this gem finds this issue interesting and the proposed solution something worth doing.

## How we use this gem (our use case)

First we generate the routes programmatically for _each locale we support_ to `app/assets/javascripts/routes/` directory, using `routes-<LOCALE>.js` as the filename. (e.g. `routes-fi.js` for Finnish routes). A snippet from our rake task looks like this:

```ruby
ROUTES_DIR = File.join('app', 'assets', 'javascripts', 'routes')
FileUtils.mkdir_p(Rails.root.join(ROUTES_DIR))

Site.each do |site|
  I18n.with_locale(site.locale) do
    file_name = File.join(ROUTES_DIR, "routes-#{I18n.locale}.js")
    JsRoutes.generate!(file_name)
  end
end
```

This is because we translate our routes to get prettier URLs in each localized version of our site. This means that e.g. our feedback path in Finnish site could be `https://venuu.fi/palaute` and on English site it could be `https://venuu.com/feedback`.

The toolchain we use to get the routes working on JS side as well:

* We use [Babel](https://babeljs.io/) to transpile this modern JavaScript to a version that can be run on IE>=9 browsers.
* We bundle our JavaScript modules together by leveraging [webpack](https://webpack.github.io/) as our module bundler.
* We build separate JS bundles for every locale we support
* We have mapped `routes` name to locale specific file with a webpack config:
  ```
  config.alias.routes = `routes/routes-${locale}`
  ```

Then we have multiple modern JS files importing various routes from that file:

```js
import { feedbackPath } from 'routes';

console.log(`The feedback path is: ${feedbackPath()}`);
```

This code ends up as as part of our `bundle/main-fi.js` file, with the final `console.log` possibly evaluating to

```js
var _routes = __webpack_require__(123 /* routes module ID */);

[...]

PR #239, opened 2018-08-20
Switch to javascript

davishmcclurg:
Hi all, thanks for the gem. It's been really useful for transitioning to webpacker. With webpacker becoming the standard way to use javascript in rails, I thought it made sense to try to convert this project to javascript so that more improvements can be made more easily. #215 explains the reasons well.

In my view, the main benefit of switching to javascript is that it opens up the possibilities for improving the webpacker integration.

There's a lot here, but most of it is converting the coffeescript files to javascript with [decaffeinate](https://github.com/decaffeinate/decaffeinate). It also adds a rollup/babel config and build process. There are a few breaking changes that will need to be worked out if you think this should be merged. I'd be happy to continue work on the branch (and additional improvements) if there's interest.


bogdan:
I also want the breaking changes to be listed. They are important to know in the first place.

spec/js_routes/amd_compatibility_spec.rb
```diff
@@ -27,12 +27,8 @@
     evaljs(JsRoutes.generate({}))
   end
 
-  it "should working from global scope" do
```
le0pard: This test should work for cases, where not "build system" (include file by `<script>` tag should be also valid)

davishmcclurg:
Breaking changes

- nested namespaces: using `namespace: 'PHM.Routes'` will not generate `window.PHM.Routes`
- globals aren't always defined (it uses a UMD module)

spec/js_routes/amd_compatibility_spec.rb
davishmcclurg: This test doesn't work right now because `window.define` is defined which prevents the global from being created when using UMD modules. I can verify that this still works in another test

le0pard:
@davishmcclurg but looks like Universal Module Definition should provide ability to browser globals without any problem.

```js
(function (root, factory) {
    if (typeof define === 'function' && define.amd) {
        // AMD
        define(['jquery'], factory);
    } else if (typeof exports === 'object') {
        // Node, CommonJS-like
        module.exports = factory(require('jquery'));
    } else {
        // Browser globals (root is window)
        root.returnExports = factory(root.jQuery);
    }
}(this, function ($) {
    //    methods
    function myFunc(){};

    //    exposed public method
    return myFunc;
}));
```
So global still should work for not defined `define` and `exports`

davishmcclurg:
@le0pard yeah, globals should work right now, just not nested ones (with periods): `namespace: 'one.two.three'`. And the previous version always defined globals (even when importing with amd, etc), which is a breaking change but I think a good one.

le0pard:
> And the previous version always defined globals (even when importing with amd, etc), which is a breaking change but I think a good one.

@davishmcclurg nope, it doesn't work in this way. It use the same approach, as UMD. 

First it check about AMD system:
https://github.com/railsware/js-routes/blob/master/lib/routes.js.coffee#L402-L407

Next it check commonjs system:
https://github.com/railsware/js-routes/blob/master/lib/routes.js.coffee#L6

And after this define namespace in `root`, which can be `exports` or `this`, where last one in browser will be `window`

https://github.com/railsware/js-routes/blob/master/lib/routes.js.coffee#L399

So old approach works even better, because support nested namespace (as I can see right now).

davishmcclurg:
It looks like it's calling `Utils.make` for the amd version, which adds a global. This test is what I'm talking about: https://github.com/railsware/js-routes/blob/master/spec/js_routes/amd_compatibility_spec.rb#L31
I don't think globals should be added when using AMD.

le0pard:
@davishmcclurg yep, it can happen by reusability of the same function. But we can change behaviour, if this needed.

davishmcclurg:
These changes won't improve the webpacker integration on their own. I'm hoping they'll open up the possibilities for future improvements. I was thinking of generating multiple builds: one that can be used as a single file and one that uses ES modules so that people can import paths separately (`import { somePath } from './js-routes'`). Using imports and modules could also help clean up the code in general by allowing separate source files.

I also believe switching to javascript will encourage more contributions. Obviously most js devs could also learn coffeescript, but fewer barriers should mean more contributors.

lib/routes-src.js
```diff
@@ -0,0 +1,482 @@
...
+  configuration: {
+    prefix: PREFIX,
+    default_url_options: DEFAULT_URL_OPTIONS,
+    special_options_key: SPECIAL_OPTIONS_KEY,
+    serializer: SERIALIZER
+  },
+
+
+  default_serializer(object, prefix = null) {
+    if (object == null) { return ""; }
```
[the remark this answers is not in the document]
davishmcclurg: `== null` is often used for checking if a variable is undefined or null: http://adripofjavascript.com/blog/drips/equals-equals-null-in-javascript.html

le0pard:
>  I was thinking of generating multiple builds: one that can be used as a single file and one that uses ES modules so that people can import paths separately (import { somePath } from './js-routes'). Using imports and modules could also help clean up the code in general by allowing separate source files.

In this case better think how this can be done, instead "let's convert everything to JS", because we can move to coffeescript 2 our route file and generate needed new files with commonjs system (and not break AMD and Browser globals). But we need to think and manage how this will be done (what will be exactly in files and is even "tree shaking" will give any benefit from this additional changes - change, measure, conclusion). Because convention not answer on this.

> I also believe switching to javascript will encourage more contributions. Obviously most js devs could also learn coffeescript, but fewer barriers should mean more contributors.

It is weak point for me.

davishmcclurg:
AMD and browser globals aren't broken in this branch. In my opinion, the current implementation is broken because it always leaks globals, but that's beside the point.

I think tree shaking would have the benefit of not having to whitelist routes, since you could generate JS for all of them and let webpack only import the ones you want.

Sounds like there's not much interest in switching to javascript, which is reasonable since coffeescript is already there and it works.

le0pard:
@davishmcclurg it will be interesting, if this could convertion solve some problems (like three shaking support, etc). But maybe @bogdan think it can be usefull. I am not against this, just still not fully see benefits.

P.S. TypeScript will be also interesting (just for fun, no big benefits too), but in this case we will loose even more contributors :)

bogdan:
As I understood, tree shaking is a technique to figure out dead code in the project. I believe it has almost no benefit for small (on the JS side) projects like js-routes: there is only one file and doubt we will ever have a large chunks of dead code - not more than 1-2 functions in practice.

As for now,  I believe coffeescript is Ok even though I won't start a project using coffee script if I do it today. 
If there is any practical problem with coffeescript implementation now: something that can not be done, lets better organize ourselves around that problem. Like no ability to import individual routes or something like that.

davishmcclurg:
I was thinking tree-shaking could be used to remove rails routes that aren't explicitly imported because they'd be dead code. That would mean you wouldn't have to whitelist routes when generating.

It looks like es modules are supported in coffeescript now: https://coffeescript.org/#modules
I'm not sure how you'd generate multiple builds, but I bet it's possible as well.

I'm going to close this since you'd prefer to try coffeescript solutions. Thanks for taking a look.

Issue #243, opened 2019-03-29
URI-encoding of route parts


orhantoy:
Hey 👋 

I noticed that route parts are not URI-encoded. I made a spec to to confirm the behavior

```diff
diff --git a/spec/js_routes/rails_routes_compatibility_spec.rb b/spec/js_routes/rails_routes_compatibility_spec.rb
index d26491f..9906ec0 100644
--- a/spec/js_routes/rails_routes_compatibility_spec.rb
+++ b/spec/js_routes/rails_routes_compatibility_spec.rb
@@ -111,6 +111,9 @@ describe JsRoutes, "compatibility with Rails"  do
     it "should support route with parameters" do
       expect(evaljs("Routes.blog_app_post_path(1)")).to eq(blog_routes.post_path(1))
     end
+    it "should support route with parameters containing symbols that need URI-encoding" do
+      expect(evaljs("Routes.blog_app_post_path('#hello')")).to eq(blog_routes.post_path('#hello'))
+    end
     it "should support root path" do
       expect(evaljs("Routes.blog_app_root_path()")).to eq(blog_routes.root_path)
     end
```

which currently fails

```
Failures:

  1) JsRoutes compatibility with Rails with rails engines should support route with parameters containing symbols that need URI-encoding
     Failure/Error: expect(evaljs("Routes.blog_app_post_path('#hello')")).to eq(blog_routes.post_path('#hello'))

       expected: "/blog/posts/%23hello"
            got: "/blog/posts/#hello"

       (compared using ==)
     # ./spec/js_routes/rails_routes_compatibility_spec.rb:115:in `block (3 levels) in <top (required)>'

Finished in 2.32 seconds (files took 0.57648 seconds to load)
160 examples, 1 failure

Failed examples:

rspec ./spec/js_routes/rails_routes_compatibility_spec.rb:114 # JsRoutes compatibility with Rails with rails engines should support route with parameters containing symbols that need URI-encoding
```

Is this an intentional choice by this lib? I would expect the behavior to match the Rails URL helpers. But I also understand that changing this could potentially break other people's use of `js-routes`, in the case where they are doing the URI-encoding themselves.

le0pard:
@orhantoy 

After some try (https://github.com/railsware/js-routes/commit/dd6c63f34ba6953bbc11dd2e0ce6cdbdbcb954e2) I get fail in another routes:

```
[...]

PR #244, opened 2019-03-30
URI-encode route parts

orhantoy:
Resolves https://github.com/railsware/js-routes/issues/243

@le0pard Thanks for replying to the issue 👍 I played a bit around with this and got the specs passing.

If you ask me, the changes I'm proposing here are breaking changes, requiring bumping the major version. This change will actually break my usage of `js-routes` because I'm doing the URI-encoding myself and then passing it into the generated route functions. That way I would end up with URI-encoding twice.

Alternatively this change could be configured with a new config option (which is disabled by default).

Happy to hear what y'all think :v:

### DIFF
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -281,7 +281,7 @@ Based on Rails routes of APP_CLASS
           value = parameters[left];
           if (value != null) {
             delete parameters[left];
-            return this.path_identifier(value);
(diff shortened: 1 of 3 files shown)


le0pard:
@bogdan need your review

bogdan:
Good catch 👍

PR #245, opened 2019-04-03
#243 Fix encoding for some url segments

le0pard:
Possible solution for https://github.com/railsware/js-routes/issues/243 issue. 

I just take regex from ActionDispatch URL Segment and provide it to our js code. Issue, that just interpolation will not work with regex, so I take `js_regex` gem, which of course introduce new dependency to js-routes.

```ruby
2.6.2 :021 > JSON.dump(ActionDispatch::Journey::Router::Utils::UriEncoder::SEGMENT)
 => "\"(?-mix:[^a-zA-Z0-9\\\\-\\\\._~!\\\\$\\u0026'\\\\(\\\\)\\\\*\\\\+,;=:@])\""
2.6.2 :022 > ActionDispatch::Journey::Router::Utils::UriEncoder::SEGMENT.inspect
 => "/[^a-zA-Z0-9\\-\\._~!\\$&'\\(\\)\\*\\+,;=:@]/"
```
in JS (by js_regex gem):

```js
UriEncoderSegmentRegex = /[\x00-\x20"#%/<>?\x5B-\x5E`{-}\x7F-\uD7FF\uE000-\uFFFF]/
```

So I need deep review from you @bogdan


bogdan:
@zhuravel @iurevych can you try this version?

lib/js_routes.rb
```diff
@@ -117,19 +118,26 @@ def generate
...
-      "SERIALIZER"          => @configuration.serializer || json(nil),
+      "DEPRECATED_GLOBBING_BEHAVIOR" => deprecated_globbing_behavior,
+
+      "APP_CLASS"                 => application.class.to_s,
+      "NAMESPACE"                 => json(@configuration.namespace),
+      "DEFAULT_URL_OPTIONS"       => json(@configuration.default_url_options),
+      "PREFIX"                    => json(@configuration.prefix),
+      "SPECIAL_OPTIONS_KEY"       => json(@configuration.special_options_key),
+      "SERIALIZER"                => @configuration.serializer || json(nil),
+      "URI_ENCODER_SEGMENT_REGEX" => (JsRegex.new(
```
bogdan: I think this new dependency can be development dependency but not runtime dependency (just like coffee script): we can compile regexps only once when the gem is "built".
           
I can even think about hardcode if the dynamic implementation would be too complex: the regexp was last updated in 2014 and I don't think it will ever be updated again. https://github.com/rails/rails/commit/5460591f0226a9d248b7b4f89186bd5553e7768f

lib/routes.js.coffee
```diff
@@ -225,6 +226,11 @@ Utils =
       else
         throw new Error("Unknown Rails node type")
 
+  encode_segment: (segment) ->
+    if UriEncoderSegmentRegex.test(segment)
```
bogdan: Why do you need this condition? Performance?

spec/js_routes/rails_routes_compatibility_spec.rb
```diff
@@ -115,8 +115,13 @@
...
       expect(evaljs("Routes.blog_app_post_path(1)")).to eq(blog_routes.post_path(1))
     end
-    it "should support route with parameters containing symbols that need URI-encoding" do
+    it "should support route with parameters containing symbols that need URI-encoding", :aggregate_failures do
       expect(evaljs("Routes.blog_app_post_path('#hello')")).to eq(blog_routes.post_path('#hello'))
+      expect(evaljs("Routes.blog_app_post_path('some param')")).to eq(blog_routes.post_path('some param'))
+    end
+    it "should support route with parameters containing symbols not need URI-encoding", :aggregate_failures do
+      expect(evaljs("Routes.blog_app_post_path(':some_id')")).to eq(blog_routes.post_path(':some_id'))
+      expect(evaljs("Routes.blog_app_post_path('.+')")).to eq(blog_routes.post_path('.+'))
```
bogdan: The routes from mounted rails engine are used here. I don't think this is good idea because the thing we actually test can be done with a simplier route which will cause less side effects from possible bugs in engines support.

zhuravel:
@bogdan Tried this version, seems to work fine.

js-routes.gemspec
```diff
@@ -31,6 +32,6 @@ Gem::Specification.new do |s|
   else
     s.add_development_dependency(%q<byebug>)
     s.add_development_dependency(%q<pry-byebug>)
-    s.add_development_dependency(%q<therubyracer>, [">= 0.12.3"])
+    s.add_development_dependency(%q<mini_racer>, [">= 0.2.4"])
```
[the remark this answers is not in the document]
le0pard: Just update V8 engine, because therubyracer use pretty old V8 and new one can speedup tests

le0pard:
@bogdan Ok, I made some fixes. 

I just copy regex from `ActionDispatch::Journey::Router::Utils::UriEncoder::SEGMENT`, because add this on runtime in our "build" process in not working as we are not knowing user rails environment. We still need `js_regex` inside user runtime (maybe exists some another case, I did not spend much time on this)

spec/spec_helper.rb
```diff
@@ -89,25 +93,14 @@ class ::App < Rails::Application
...
-        value.to_h.map do |k,v|
-          [k, inspectify(v)]
-        end.to_h
-      when String, nil, Integer, FalseClass, TrueClass
-        value
-      else
-        raise "wtf #{value.class}?"
+    if defined?(JRUBY_VERSION)
+      jscontext[:log] = lambda do |context, value|
+        puts value
```
bogdan: Is it just working now and we don't need to format manually?

js-routes.gemspec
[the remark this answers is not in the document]
bogdan: That's a little bad this change is merged to this hotpatch... hotpatch, you know, should be fast...

spec/js_routes/rails_routes_compatibility_spec.rb
```diff
@@ -14,15 +14,15 @@
     expect(evaljs("Routes.inbox_path(1)")).to eq(test_routes.inbox_path(1))
   end
 
-  it "should raise error if required argument is not passed" do
+  it "should raise error if required argument is not passed", :aggregate_failures do
     expect { evaljs("Routes.thing_path()") }
-      .to raise_error('Route parameter missing: id')
+      .to raise_error(/parameter missing: id/)
```
bogdan: Why is it changed?

spec/spec_helper.rb
le0pard: @bogdan yes, it is benefit from mini_racer - it is doing converting of objects from JS-to-Ruby itself.

spec/js_routes/rails_routes_compatibility_spec.rb
le0pard: Because mini_racer return `Error: Route parameter missing: id` as error message. But for jruby therubyrhino gem will return message `Route parameter missing: id`

le0pard: Made `/Route parameter missing .. /` regex

bogdan:
v1.4.6 ❤️

iurevych:
Thank you guys 👍

le0pard:
Thanks @bogdan

PR #266, opened 2020-11-30
Add assign helper

tnandate:
`Object.assign()` does not work with Internet Explorer.
So I would like to add an assign helper to make it work in the same way as `indexOf`.
copied from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

### DIFF
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -467,6 +467,36 @@ Based on Rails RAILS_VERSION routes of APP_CLASS
       }
       return result;
     },
+    assign: function() {
+      var sources, target;
+      target = arguments[0], sources = 2 <= arguments.length ? slice.call(arguments, 1) : [];
+      if (typeof Object.assign === 'function') {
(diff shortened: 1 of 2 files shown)


le0pard:
Nobody care about IE, even Microsoft itself - https://techcommunity.microsoft.com/t5/microsoft-365-blog/microsoft-365-apps-say-farewell-to-internet-explorer-11-and/ba-p/1591666 . 

If you need support it, use core.js polifils https://alexbogovich.com/blog/core-js/

le0pard:
BTW, it is good idea to cleanup old helpers, which was added only for IE (like `indexOf`). What do you think @bogdan ?

bogdan:
> BTW, it is good idea to cleanup old helpers, which was added only for IE (like indexOf). What do you think @bogdan ?


Agree, I don't think anyone cares now and we should not too.

tnandate:
OK. This pull request will be closed.

PR #269, opened 2020-12-07
Add `module.exports` for node environments

### LINKED ISSUE #268: Require returns empty object in node environments
Upon trying to set up javascript testing, I noticed when the routes file is required in node environments it returns an empty object.

```
> const routes = require('./routes')
undefined
undefined
> routes
{}
```

Tested with node 10.x and 12.x

Also! This is a webpacker setup, so that may additionally be a factor.

jondavidjohn:
Fixes #268 

Noticed that when namespace ends up `null` (like in `Webpacker` environments), and you try and import the routes in a node environment (for running specs against your components in Jest), the require comes back as an empty object because the routes never get assigned to root.

This is likely due to 7c654aa, specifically the detection of `Webpacker` and marking the namespace to `nil`.

When in these types of environments, we actually want to use `module.exports`.

Not super sure how to compile the coffeescript into the js template, but this should fix the issue.


bogdan:
@jondavidjohn is there any way to test this?

jondavidjohn:
@bogdan Added a spec, similar to the AMD specs that exist.

jondavidjohn:
Additionally, I've tested it manually in a project and can confirm it works as expected.

bogdan:
Do you have any idea if `define` and `module` can both be defined and if we should handle a case like this by preferring one over the other instead of assigning both?

jondavidjohn:
@bogdan I think it's probably a good idea to make an explicit choice between:

1. AMD
2. CommonJS
3. Global Browser (namespaced)

I did have to remove a few specs that asserted it is still global, even when using AMD, which didn't make sense to me.

Otherwise I think this is good to go.

jondavidjohn:
@bogdan Looks like these changes didn't make it in `1.4.12`?

https://github.com/railsware/js-routes/compare/v1.4.11...v1.4.12

bogdan:
👍  Released 1.4.13

lib/routes.js
```diff
@@ -513,6 +510,10 @@ Based on Rails RAILS_VERSION routes of APP_CLASS
     define([], function() {
       return result;
     });
+  } else if (typeof module !== "undefined" && module !== null) {
```
davegudge, line 513: I've been using js-routes with webpacker since we migrated our javascript from the assest pipeline.

Following the [change](https://github.com/railsware/js-routes/commit/7c654aa0c3055e3cf48a87050fad96a8befd9bad#diff-e04a6967ae1704362c5f4fccb26e7b29296144d2bc5f3c785974d71457418c8aR14) in 7c654aa0c3, I had to specify an initaliser to redefine the namespace (to keep our setup working):

```ruby
JsRoutes.setup do |config|
  config.namespace = "Routes"
end
```

I believe the namespace is no longer honoured since this change, and I'm getting an error of:

Screenshot 2020-12-18 at 13 01 21

Perhaps I'm not importing js-routes in the way that was intended? 

`app/javascript/javascripts/index.coffee.erb`
```coffee
# js-routes
# https://github.com/railsware/js-routes/issues/234#issuecomment-380794815
import Routes from "./routes.js.erb";
# Make the routes available globally
window.Routes = Routes;
```

Where `routes.js.erb` is defined as:

```erb
/* rails-erb-loader-dependencies ../config/routes */
var routes = {};

(function() {
    <%= JsRoutes.generate %>
}.call(routes));

export default routes.Routes;

```

bogdan: Your way of import is pretty odd. I think you just need the following:

```
const routes = <%= JsRoutes.generate %>
export default routes;
```

Also `namespace` option is not needed in this case.

davegudge: Thanks for the information. 

I was getting the same error using the suggestion: 

```
TypeError: Cannot assign to read only property 'exports' of object '#<Object>' error
```

But it seems to work if I used the CommonJS syntax, e.g:

```
const routes = <%= JsRoutes.generate %>;
module.exports = routes;
```

Issue #281, opened 2021-07-13
TypeError: Cannot set property 'NameSpace' of undefined at version 2.0.7


tashirosota:
Ruby: 2.6.6
Rails: 6.1.3
Raise TypeError: Cannot set property 'NameSpace' of undefined when I tried for update to 2.0.7 from 1.4.14

```rb
# setup & generate
JsRoutes.setup do |config|
  config.module_type = 'UMD'
  config.namespace = 'RailsRoutes'
  config.documentation = false
end
JsRoutes.generate! path, options
```

 - Where
```js
// 2.0.7
// Raise in rails-routes.js
debugger
Utils.namespace(Root, "RailsRoutes", result);

// In console
$ Root
 => undefined
$ thas
 => undefined

// 1.4.14
// Raise in rails-routes.js
if (typeof define === "function" && define.amd) {
  define([], function() {
    return result;
  });
} else if (typeof module !== "undefined" && module !== null) {
  try {
    module.exports = result;
  } catch (error1) {
    error = error1;
    if (error.name !== 'TypeError') {
      throw error;
    }
  }
} else {
  debugger // Can not reach this line
  Utils.namespace(Root, "RailsRoutes", result);
}
```

Is this implementation out of place?
```js
if (typeof define === "function" && define.amd) {
  define([], function() {
    return result;
  });
} else if (typeof module !== "undefined" && module !== null) {
  try {
    module.exports = result;
  } catch (error1) {
    error = error1;
    if (error.name !== 'TypeError') {
      throw error;
    }
  }
} else {
  Utils.namespace(Root, "RailsRoutes", result);
}
```

bogdan:
If you are using `UMD` module type, you'd probably need to remove `namespace` option. This option is only valuable if `module_type` is `nil`.

tashirosota:
@bogdan 
Thanks！！！I resolved！！

PR #282, opened 2021-07-13
fix/utils_namespace

tashirosota:
https://github.com/railsware/js-routes/issues/281


bogdan:
I think the better fix of #281 would be to disallow `namespace` option when module type is `UMD`. See: https://github.com/railsware/js-routes/issues/281#issuecomment-882723889

PR #287, opened 2021-10-19
Add safe check for 'window' to support server rendering

AlexMeah:
Without this check an exception is raised when server rendering. This is masked by `evaljs("var window = this;", {force: true})` is spec_helper.rb


le0pard:
```js
node                                
Welcome to Node.js v16.9.1.
Type ".help" for more information.
> typeof window !== "undefined"
false
```

Why `isBroswer` will raise error on node.js @AlexMeah ?

le0pard:
Looks ok. Your move @bogdan

AlexMeah:
@bogdan Please can you do a new release

bogdan:
Released 2.1.2 :+1:

PR #306, opened 2023-05-16
Add post_processor to customize the output of each route

vovanmozg:
Sometimes we need to customize the output of each individual route.

I was faced with the task of typing routes when used with react. And I couldn't use dynamic generated routes with toString(). This method returns `string` type for any helper. It would be great for me if it was possible to export patterns as literals

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -302,7 +302,9 @@ Options to configure JavaScript file generator. These options are only available
   * Default: `Rails.application`
 * `file` - a file location where generated routes are stored
   * Default: `app/javascript/routes.js` if setup with Webpacker, otherwise `app/assets/javascripts/routes.js` if setup with Sprockets.
-
+* `post_processor` - a proc that takes a string and returns a string. It is called after the each route is generated and can be used to modify the generated code.
+  * Default: `nil`
+  * Example: `->(code) { code + addition }`
 <div id="formatter-options"></div>
 
 #### Formatter Options
--- a/lib/js_routes/configuration.rb
+++ b/lib/js_routes/configuration.rb
@@ -17,6 +17,7 @@ module JsRoutes
       application: -> { Rails.application },
       module_type: 'ESM',
       documentation: true,
+      post_processor: nil,
     } #:nodoc:
 
     attr_accessor(*DEFAULTS.keys)
@@ -32,9 +33,10 @@ module JsRoutes
         raise "Provide attributes or block"
       end
       tap(&block) if block
+      no_instant_call_attributes = [:post_processor]
       if attributes
         attributes.each do |attribute, value|
-          value = value.call if value.is_a?(Proc)
+          value = value.call if value.is_a?(Proc) && !no_instant_call_attributes.include?(attribute)
           send(:"#{attribute}=", value)
         end
       end
--- a/lib/js_routes/instance.rb
+++ b/lib/js_routes/instance.rb
@@ -112,15 +112,21 @@ module JsRoutes
           "", name,
           @configuration.dts? ?
           "RouterExposedMethods['#{name}']" :
-          "__jsr.#{name}"
+          "__jsr.#{name}",
+          nil
         ]
       end
     end
 
+    def post_process(content, route)
+      @configuration.post_processor ? @configuration.post_processor.call(content, route) : content
+    end
+
     def routes_export
       return "" unless @configuration.modern?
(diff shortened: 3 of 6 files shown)


bogdan:
Can you give some live examples of how `post_processor` would look like in your app?

vovanmozg:
I regret that I did not motivate PR well enough.

Consider the following example. 

I have React Router routes description in routes.ts file:
```typescript
import React from "react";
import ReactDOM from "react-dom/client";
import { createBrowserRouter, RouterProvider } from "react-router-dom";

const router = createBrowserRouter([
  {
    path: "/settings",
    element: <div>...</div>
  },
  {
    path: "/profile",
    element: <div>...</div>
  },
]);

ReactDOM.createRoot(document.getElementById("root")).render(
  <React.StrictMode>
    <RouterProvider router={router} />
  </React.StrictMode>
);
```
In this case type of `path` property is union of two literals "/settings" | "/profile".

I can use .toString() to define path: 
```
const router = createBrowserRouter([
  {
    path: Routes.settings_path.toString().replace(".:format", ""),
    element: <div>...</div>
  }
]);
```
But in this case the type of `path` will become just a string. But I know for sure that the path is only limited to a certain subset. And I want to use this knowledge for type checking

vovanmozg:
Yes, the solution looks pretty hacky. I don't like that the post_processor has to deal with an already formed string. Perhaps it should receive not a string, but an array of components [comment, name, body].

bogdan:
I am not sure how your solution is going to work. 
At first, `settings_path()` and `profile_path()` will always return a string because it is dynamically constructed.
So your only option to always use `/settings` and `/profile` when you would want to actually use the path:

``` typescript
const settings_path = (): string => "/settings";
const path: "/settings" | "/profile" = settings_path();
```

So, even if you can dynamically define a router using js-routes helpers, you would have to hardcode it later. So, what is the point of avoiding hardcode with js-routes in the first place?

Other thing is route parameters like `/posts/:id`. I don't see any way to ensure certain variable can only be assigned a result of `post_path(id)` but not `message_path(id)`. Do you plan to support parametrized routes in the first place?

vovanmozg:
Lets look at test one more time
```ruby
Proc.new do |content, route|
  if route
    # This may be a more complex replacement.
    pattern = route.path.spec.to_s.gsub('(.:format)', '').gsub('.:format', '')
    content + %Q[,\n#{route.name}_pattern: '#{pattern}']
  else
    content
  end
end
```
https://github.com/railsware/js-routes/pull/306/files#diff-544bb65018327fd52b3c481ebef39d767ee5e876a08a89fe2037901b82f3f91d

It generates the code:
```javascript
/**  
* Generates rails route to  
* /inboxes/:inbox_id/messages(.:format)  
* @param {any} inbox_id  
* @param {object | undefined} options  
* @returns {string} route path  
*/  
inbox_messages_path: __jsr.r({"inbox_id":{"r":true},"format":{}}, [2,[7,"/"],[2,[6,"inboxes"],[2,[7,"/"],[2,[3,"inbox_id"],[2,[7,"/"],[2,[6,"messages"],[1,[2,[8,"."],[3,"format"]]]]]]]]]),  
inbox_messages_pattern: '/inboxes/:inbox_id/messages',
```

Then I can import and use `inbox_messages_pattern` to define React Router item:
```javascript
...
import { inbox_messages_pattern } from 'routes';
...
const router = createBrowserRouter([
  {
    path: inbox_messages_pattern,
    element: <div>...</div>
  }
]);

```

And I don't have to hardcode literal (`/inboxes/:inbox_id/messages`) this place. It is way to avoid routes duplication.

bogdan:
So what we can do. First, rename `toString()` to `spec()` because of some limitation over the return type of `toString()` (honestly, naming it `toString()` always was a bad idea).

``` typescript
type RouteHelperExtras<T extends string = string> = {
  ....
  spec(): T;
}
```

Make sure we generate `RouteHelperExtras<"/settings">` with appropriate spec here: https://github.com/railsware/js-routes/blob/master/lib/js_routes/route.rb#L43


The only problem of such approach: it will not strip optional parameters like `:format` automatically.

bogdan:
I decided that it is too much of an overhead and isn't applicable to all use cases to accept it to the core. So closing this one until a better solution will be found.

vovanmozg:
I agree with you. The idea was underdeveloped

PR #308, opened 2023-07-26
Add #__PURE__ to __jsr.r call in ESM output only, update tests.

### LINKED ISSUE #307: ESM Tree shaking not working because __jsr.r() is potentially impure
Thank you so much for maintaining this package! We're using it in my company and are moving to the ESM format in a Webpack setup.

## Problem

I noticed that even though each route is exported individually, the way they are defined prevents Webpack (and other bundlers) from being able to tree-shake unused exports. 

To explain: each export is the result of a function call

```javascript
export const accept_user_invitation_url = __jsr.r({"format":{}}, [2,[7,"/"],[2,[6,"users"],[2,[7,"/"],[2,[6,"invitation"],[2,[7,"/"],[2,[6,"accept"],[1,[2,[8,"."],[3,"format"]]]]]]]]], true);
```

and Webpack can't prove that the function call doesn't have a side-effect (like modifying module/global variables), so it will decide not to remove it, for safety. The way libraries typically deal with this is by adding a `/*#__PURE__*/` [comment](https://webpack.js.org/guides/tree-shaking/#mark-a-function-call-as-side-effect-free) right before the call, to inform the minifier that the call is safe to remove if possible.

## Reproduction

I've created a minimal reproduction [here](https://github.com/joshwilsonvu/js-routes-treeshaking). It contains two nearly identical directories, each with a `src/index.js` that uses only one of the sixteen routes defined in `src/routes.js` (a partial copy of my codebase's much larger file). In each directory, `dist/main.js` is generated from running `yarn build`.

 The `without-pure` directory's `src/routes.js` is what `js-routes` generates today; the `without-pure` directory's `src/routes.js` has been modified to include a `/*#__PURE__*/` comment right before each `__jsr.r()` call. This is the only difference.

In `without-pure`, tree shaking doesn't work; `without-pure/dist/main.js` contains code defining all the routes. Scroll down and you'll see many `r.r()` calls (minified from `__jsr.r`).

In `with-pure`, tree shaking works; `with-pure/dist/main.js` has only one `r.r()` call, and half as many total lines.

## Suggested solution

I think we can add the `/*#__PURE__*/` comment simply by

```diff
    def routes_export
      return "" unless @configuration.modern?
      [*static_exports, *routes_list].map do |comment, name, body|
-       "#{comment}export const #{name}#{export_separator}#{body};\n\n"
+       "#{comment}export const #{name}#{export_separator}/*#__PURE__*/ #{body};\n\n"
      end.join
    end
```

It should only apply to the ESM format. This could break a few tests in `spec/js_routes/module_types/esm_spec.rb`, but those are fixed by adding the `/*#__PURE__*/` comment as appropriate.

I would be happy to submit a pull request to do this. Please let me know if you have any questions or concerns! Thanks again.

joshwilsonvu:
Closes #307.

I added the `#__PURE__` comment to `Route#body` like you suggested and that worked great! Just needed to make a few small test changes. 

It would be nice to add a test to ensure Webpack does tree-shake the output, I'm just not sure how I'd do that. But given the results of my example repo, I think we can be pretty confident that this change will work.

Thanks in advance for the review and feel free to make any changes!

### DIFF
--- a/lib/js_routes/route.rb
+++ b/lib/js_routes/route.rb
@@ -38,6 +38,7 @@ module JsRoutes
       else
         # For tree-shaking ESM, add a #__PURE__ comment informing Webpack/minifiers that the call to `__jsr.r`
         # has no side-effects (e.g. modifying global variables) and is safe to remove when unused.
+        # https://webpack.js.org/guides/tree-shaking/#clarifying-tree-shaking-and-sideeffects
         pure_comment = @configuration.esm? ? '/*#__PURE__*/ ' : ''
         "#{pure_comment}__jsr.r(#{arguments(absolute).map{|a| json(a)}.join(', ')})"
       end
--- a/lib/js_routes/route.rb
+++ b/lib/js_routes/route.rb
@@ -33,8 +33,14 @@ module JsRoutes
     end
 
     def body(absolute)
-      @configuration.dts? ?
-        definition_body : "__jsr.r(#{arguments(absolute).map{|a| json(a)}.join(', ')})"
+      if @configuration.dts?
+        definition_body
+      else
+        # For tree-shaking ESM, add a #__PURE__ comment informing Webpack/minifiers that the call to `__jsr.r`
+        # has no side-effects (e.g. modifying global variables) and is safe to remove when unused.
+        pure_comment = @configuration.esm? ? '/*#__PURE__*/ ' : ''
+        "#{pure_comment}__jsr.r(#{arguments(absolute).map{|a| json(a)}.join(', ')})"
+      end
     end
 
     def definition_body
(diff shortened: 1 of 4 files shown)


lib/js_routes/route.rb
```diff
@@ -33,8 +33,14 @@ def helper_types
     end
 
     def body(absolute)
-      @configuration.dts? ?
-        definition_body : "__jsr.r(#{arguments(absolute).map{|a| json(a)}.join(', ')})"
+      if @configuration.dts?
+        definition_body
+      else
+        # For tree-shaking ESM, add a #__PURE__ comment informing Webpack/minifiers that the call to `__jsr.r`
+        # has no side-effects (e.g. modifying global variables) and is safe to remove when unused.
```
bogdan, line 40:
```suggestion
        # has no side-effects (e.g. modifying global variables) and is safe to remove when unused.
        # https://webpack.js.org/guides/tree-shaking/#clarifying-tree-shaking-and-sideeffects
```

bogdan:
Looks good to me. Please also update readme and remove the recommendation to import individual routes for tree shaking.

joshwilsonvu:
Done!

PR #310, opened 2023-08-02
fix: rename `File.exists?` to `File.exist?`

arthurgeek:
The deprecated `File.exists?` alias was removed on Ruby 3.2.0.

`File.exist?` works on all versions.

ref: https://www.ruby-lang.org/en/news/2022/12/25/ruby-3-2-0-released/

### DIFF
--- a/lib/js_routes/generators/base.rb
+++ b/lib/js_routes/generators/base.rb
@@ -10,7 +10,7 @@ class JsRoutes::Generators::Base < Rails::Generators::Base
       "app/javascript/packs/application.js",
       "app/javascript/controllers/application.js",
     ].find do |path|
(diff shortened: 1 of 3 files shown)


sbscully:
Would it be possible to release a new gem version with this fix included? I'm tracking master in my Gemfile for now but I would like to pin the version.

bogdan:
Version 2.2.8 includes this change.

m-shimojo:
2.2.8 has not been released to rubygems yet, right?
https://rubygems.org/gems/js-routes

Am I missing something?

bogdan:
Sorry, you are right. Released 2.2.8.

m-shimojo:
Thanks!

PR #315, opened 2024-09-08
Fix broken configuration when module_type is Nil and namespace is being used.

rapito:
Hi!

I have a simple project that doesn't use any import/export modules and is still using sprockets. 
On latest version, if I have this configuration:

```ruby
JsRoutes.setup do |c|
  c.module_type = nil
  c.namespace = 'Routes'
end
```

It breaks like this when I load any endpoint:

image

After reviewing, I noticed that 'DTS' was being used by default causing this clash. I solved it by monkey patching it on my project, but I thought I would suggest this modification to support the intended behavior of prioritizing DTS only if the configuration_type has not been set.

### DIFF
--- a/lib/js_routes.rb
+++ b/lib/js_routes.rb
@@ -57,7 +57,7 @@ module JsRoutes
 
     sig { returns(T.nilable(String)) }
     def default_module_type
-      'DTS' if configuration.module_type
+      'DTS' if configuration.module_type && configuration.module_type != 'NIL'
     end
 
 
--- a/lib/js_routes.rb
+++ b/lib/js_routes.rb
@@ -41,19 +41,26 @@ module JsRoutes
 
     sig { params(opts: T.untyped).returns(String) }
     def definitions(**opts)
-      generate(**opts, module_type: 'DTS',)
+      generate(**opts, module_type: default_module_type,)
     end
 
     sig { params(file_name: FileName, opts: T.untyped).void }
     def definitions!(file_name = nil, **opts)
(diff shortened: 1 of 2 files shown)


bogdan:
I am not sure TS definitions can be used in case when module type isn't ESM. Are you really using definitions in legacy setup? Do they work as intended helping you typecheck or autocomplete?

The problem way come from here if you are using the middleware:
https://github.com/railsware/js-routes/blob/master/lib/js_routes/middleware.rb#L41
Which generates definition no matter module type, while it should only happen for ESM (aka `JsRoutes.configuration.modern?`).

rapito:
> I am not sure TS definitions can be used in case when module type isn't ESM. Are you really using definitions in legacy setup? Do they work as intended helping you typecheck or autocomplete?

I don't mean to use ts nor definitions at all, I was just updating my project that uses the JSRoutes gem without any import/export module.

> The problem way come from here if you are using the middleware: https://github.com/railsware/js-routes/blob/master/lib/js_routes/middleware.rb#L41 Which generates definition no matter module type, while it should only happen for ESM (aka `JsRoutes.configuration.modern?`).

That is correct, it is part of the stacktrace. Do you suggest just refactoring my PR to add an if block there so it looks something like this:
```ruby
    def regenerate
      JsRoutes.generate!
      JsRoutes.definitions! if configuration.moden?
    end
```

bogdan:
Yeah, I think this is more logical to do....  `JsRoutes.definition` should always use DTS model.

rapito:
> Yeah, I think this is more logical to do.... `JsRoutes.definition` should always use DTS module.

Thanks, I'll retest this tonight and resubmit.

PR #341, opened 2026-03-03
Add `package` and `package_file` options

### LINKED ISSUE #339: [feature request] add public method to generate __jsr object
## Background

I am customizing the output of js-routes because of some customizations in our routing system and to split the path helpers up into multiple files. I am using `JsRoutes::Route` to generate the path helper but there is no built in way to generate the `__jsr` object without also generating the path helpers. I am using this workaround at the moment but it would be cleaner if I could just call a method.

```rb
# frozen_string_literal: true

# We want the js-routes JavaScript utils but need to customize the output of the JavaScript path helpers.
# To force js-routes to generate the JavaScript utils but not the JavaScript path helpers
# we can pass a fake `Rails::Application` object with empty `named_routes`.
# We then customize the output of the JavaScript path helpers in `lib/gitlab/js_routes.rb`
class EmptyRoutesApplication
  Routes = Struct.new(:named_routes)

  NamedRoutes = Struct.new do
    def to_h
      {}
    end
  end

  def routes
    Routes.new(NamedRoutes.new)
  end

  def reload_routes_unless_loaded; end

  def reload_routes!; end

  def is_a?(klass)
    klass == Rails::Application
  end
end

JsRoutes.setup do |c|
  c.application = -> do
    EmptyRoutesApplication.new
  end
end
```

## Proposal

Add a public method such as `generate_jsr` to `lib/js_routes/instance.rb`. This will allow more custom setups where you can fully customize the output.

phegman:
Closes https://github.com/railsware/js-routes/issues/339

Adds more flexibility to split routes into multiple files with a shared package imported between them. This can improve tree shaking for codebases with a large number of routes. See example in docs (in this PR) and https://github.com/railsware/js-routes/issues/339#issuecomment-3783525416 for more context

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -240,6 +240,29 @@ which will cause any `JsRoutes` instance to generate defintions instead of route
 
 <div id="sprockets"></div>
 
+#### Splitting routes into multiple files with shared package (`ESM` only)
+
+Some setups may benefit from splitting routes into multiple files with the core js-routes utils shared between these files. This can be helpful for codebases that have a large number of routes and can improve tree shaking. This is only available when the `module_type` is set to `ESM`.
+
+```ruby
+class AdvancedJsRoutesMiddleware < JsRoutes::Middleware
+  def regenerate
+    JsRoutes.generate_package!
+
+    JsRoutes.generate!(
+      "admin_routes.js",
+      include: /^admin_/,
+      package: './routes_core.js'
+    )
+    JsRoutes.generate!(
+      "api_routes.js",
+      include: /^api_/,
+      package: './routes_core.js'
+    )
+  end
+end
+```
+
 ### Sprockets (Deprecated)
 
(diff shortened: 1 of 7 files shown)


bogdan:
> @bogdan sorry for the delay on this, I've was on my honeymoon with a strict no coding policy 😆.

Oh 100% there with you. 👍 Make love is more important than ever!

I would need some time to digest this draft to make sure we don't make any design choices that cause long term backward compatibility issues when we would want to improve things. I will try to be quick.

bogdan (changes requested):
After thinking about it for entire week I didn't find any reason why this is not the step the right direction or any compatibility problems it can cause in the future.
Good job!

bogdan:
@phegman as you are probably more busy now then I am, I decided to finalized it. I welcome your feedback even post-merge.

PR #346, opened 2026-05-28
Add `include_undefined_query_parameters` config flag

### LINKED ISSUE #345: `undefined` values treated as `nil` in Rails rather than omitted
I ran into an issue when upgrading to Rails 8.1 where previous empty strings were becoming undefined.

I know that there is the flag to use the deprecated < 8.1 behavior with the params, but at first glance it seems to me that an option to have the default serializer skip them entirely might be more in line with what JS tends to consider "undefined" as -- not provided or unset.

I know I can configure my own serializer, but I wanted to see if there was interest in my fork mod adding a simpler config flag to opt into skipping those by default as a timesaver.

You can see what I mean in this PR in our repo at the moment https://github.com/user-interviews/js-routes/pull/2 but happy to open one here if not opposed.

jbaxendale-ut:
JavaScript uses `undefined` to mean “not provided,” but js-routes previously treated `undefined` query object values the same as `null`. With Rails 8.1’s `nil` parameter handling, that means `undefined` values can become present `nil` params, serialized as bare keys like `?foo`.

This PR adds a transitional config option, `include_undefined_query_parameters`, for the default serializer:
  - `unset`: preserve legacy behavior and warn so upgrading apps choose deliberately
  - `true`: include `undefined` query properties as Rails nil
  - `false`: omit object properties whose value is `undefined`

Newly generated initializers set `include_undefined_query_parameters` to `false`. Explicit `null` continues to mean Rails `nil` and keeps the Rails-version-specific serialization behavior, including Rails 8.1 bare-key params.

The PR also updates Rails version checks to use `Gem::Version`, improves deprecation handling, and expands appraisal/CI coverage for Rails 8.0 and Rails 8.1 and Ruby 4.0

closes #345


lib/routes.ts
```diff
@@ -80,6 +81,7 @@ declare const RubyVariables: {
   PREFIX: string;
   DEPRECATED_FALSE_PARAMETER_BEHAVIOR: boolean;
   DEPRECATED_NIL_QUERY_PARAMETER_BEHAVIOR: boolean;
+  OMIT_UNDEFINED_QUERY_PARAMETERS: boolean | null;
```
bogdan: At this stage it should be passed to js as non-nullable `boolean` - ruby should have final decision on that.

spec/js_routes/generators/middleware_spec.rb
```diff
@@ -6,4 +6,9 @@
     expect(JsRoutes::Generators::Middleware.source_root).to eq(gem_root.join('lib/templates').to_s)
   end
 
+  it "sets the undefined query parameter migration option in the generated initializer" do
+    expect(File.read(gem_root.join('lib/templates/initializer.rb'))).to include(
+      "c.omit_undefined_query_parameters = true"
+    )
+  end
```
bogdan: Feels useless, lets delete it.

spec/js_routes/default_serializer_spec.rb
```diff
@@ -28,4 +31,86 @@
...
+
+  def evaluate_routes_for(rails_version:)
+    allow(Rails).to receive(:version).and_return(rails_version)
+    JsRoutes::Utils.deprecator.silence do
+      evallib(**options)
+    end
+  end
+
+  def rails_nil_query_parameter(key)
+    Gem::Version.new(Rails.version) < Gem::Version.new("8.1.0") ? "#{key}=" : key
```
bogdan: I think we can avoid this hack by having rails 8.1 in build matrix and using rails serializer to evaluate the result of js serializer.

spec/js_routes/default_serializer_spec.rb
```diff
@@ -28,4 +31,86 @@
...
+
+        expectjs("Routes.serialize({a: null, b: undefined})").to eq("a")
+      end
+    end
+  end
+
+  def evaluate_routes_for(rails_version:)
+    allow(Rails).to receive(:version).and_return(rails_version)
+    JsRoutes::Utils.deprecator.silence do
+      evallib(**options)
```
bogdan: I think we just need to always specify the new config parameter to avoid warning in tests. Maybe even inside `evallib` if not specifed from outside.

spec/js_routes/rails_routes_compatibility_spec.rb
```diff
@@ -294,6 +294,11 @@
         expectjs("Routes.thing_path(5, {optional_id: undefined})").to eq(test_routes.thing_path(5, :optional_id => nil))
       end
 
+      it "should treat undefined as non-given optional part with undefined query omission enabled" do
+        evallib(module_type: nil, namespace: 'Routes', omit_undefined_query_parameters: true)
+        expectjs("Routes.thing_path(5, {optional_id: undefined})").to eq(test_routes.thing_path(5, :optional_id => nil))
+      end
+
```
bogdan: Lets move this one to migration test.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Pending
 
 * Support `config.javascript_path` Rails configuration. Fixes [#344](https://github.com/railsware/js-routes/issues/344).
+* Add `omit_undefined_query_parameters` migration option for omitting `undefined` query parameters.
```
bogdan: Add more details on the migration and recommendations, explain the warning.

bogdan (changes requested):
Great job, looking forward to work it out!

.github/workflows/ci.yml
```diff
@@ -25,6 +25,8 @@ jobs:
             gemfile: gemfiles/rails70_sprockets_4.gemfile
           - ruby: "3.3"
             gemfile: gemfiles/rails80_sprockets_4.gemfile
+          - ruby: "3.4"
```
jbaxendale-ut: figured may as well test 8.1 against ruby ~3.4~ 4.0

lib/js_routes/instance.rb
```diff
@@ -94,12 +94,16 @@ def remove!
...
       prefix = @configuration.prefix
       prefix = prefix.call if prefix.is_a?(Proc)
+      rails_version = JsRoutes::Utils.rails_version
       {
         'ROUTES_OBJECT'       => routes_object,
-        'DEPRECATED_FALSE_PARAMETER_BEHAVIOR' => Rails.version < '7.0.0',
-        'DEPRECATED_NIL_QUERY_PARAMETER_BEHAVIOR' => Rails.version < '8.1.0',
+        'DEPRECATED_FALSE_PARAMETER_BEHAVIOR' => rails_version < Gem::Version.new('7.0.0'),
+        'DEPRECATED_NIL_QUERY_PARAMETER_BEHAVIOR' => rails_version < Gem::Version.new('8.1.0'),
+        'OMIT_UNDEFINED_QUERY_PARAMETERS' => json(@configuration.omit_undefined_query_parameters == true),
```
jbaxendale-ut: addresses the JS side always being true or false

lib/routes.js
```diff
@@ -132,14 +133,21 @@ RubyVariables.WRAPPER(() => {
             const result = [];
             if (this.is_array(value)) {
                 for (const element of value) {
-                    result.push(this.default_serializer(element, prefix + "[]"));
+                    const subvalue = this.default_serializer(element, prefix + "[]");
```
jbaxendale-ut: Did a separate review with Codex of the changes against main and it came up with this:

> • The new undefined-omission behavior leaves empty serialized array entries in some nested-array cases, producing malformed query strings with stray separators.
> 
>   Review comment:
> 
>   - [P2] Filter empty array serializations after omitting undefined — /js-routes/lib/routes.ts:269-273
>     When omit_undefined_query_parameters is enabled, an object inside an array can now serialize to an empty string if all of its properties are undefined; the array branch still pushes that empty sub-result, so Routes.serialize({a: [{b: undefined}, {c: 1}]}) returns &a%5B%5D%5Bc%5D=1 with a leading empty
>     parameter. Filter empty sub-serializations in the array path the same way the object path does.

On whether this bug was caused by this branch:

>   The underlying flaw was pre-existing: the array branch already pushed empty serialized fragments unconditionally. So even on main, an empty object inside an array could produce a stray separator.
> 
>   But the specific reviewed bug is caused/exposed by this branch. Before omit_undefined_query_parameters, this:
> 
>   { b: undefined }
> 
>   did not serialize to empty; it serialized like Rails nil. This branch added behavior where that object can become empty after omitting b, and the old array behavior then turns that empty fragment into a leading &.
> 
>   So I’d describe it as: pre-existing latent bug in array serialization, newly exposed by this branch’s undefined-omission behavior. It’s appropriate to fix in this branch.

On checking backward compatibility concerns:
> There’s already a Rails compatibility spec for empty array/object route params, which is useful context: empty structures serialize to no query params already, but the old array branch could still leak separators when an empty serialized child sat next to a non-empty one. I’ll inspect that compatibility expectation before giving the deploy recommendation.

And whether it should be limited to just the new config:
> It’s generally safe to deploy across the board. I would not limit this fix to omit_undefined_query_parameters === true.
> 
>   Reason: the fix does not change how any real value serializes. It only prevents an empty serialized child fragment from being included in the internal result.join("&").
> 
>   This is a malformed-output cleanup:
> 
>   ["", "a%5B%5D%5Bc%5D=1"].join("&")
>   // before: "&a%5B%5D%5Bc%5D=1"
> 
>   ["a%5B%5D%5Bc%5D=1"].join("&")
>   // after: "a%5B%5D%5Bc%5D=1"
> 
>   The only consumers who could observe a change are ones depending on stray separators like leading &, trailing &, or doubled && from empty nested array entries. That’s not a meaningful compatibility contract, and it is inconsistent with the object branch, which already filters empty sub-serializations.
> 
>   Limiting it to the config flag would be more complex and less coherent:
> 
>   if (this.configuration.omit_undefined_query_parameters && subvalue.length === 0) {
>     continue;
>   }
>   result.push(subvalue);
> 
>   That preserves malformed separators for other cases, like arrays containing empty objects next to real values. Since Rails compatibility already treats empty arrays/objects as producing no query params, the cleaner behavior is: any recursive serialization that returns "" should not contribute an array item to the joined query string.
> 
>   So my recommendation: keep the current unconditional fix. It’s small, consistent with existing object behavior, and the risk is limited to removing invalid separator noise.

lib/routes.ts
```diff
@@ -49,6 +49,7 @@ type RouteHelpers = Collection<RouteHelper>;
 type Configuration = {
   prefix: string;
   default_url_options: RouteParameters;
+  omit_undefined_query_parameters: boolean;
```
jbaxendale-ut: no longer nullable

lib/routes.ts
```diff
@@ -256,12 +260,21 @@ RubyVariables.WRAPPER((): RouterExposedMethods => {
       const result: string[] = [];
       if (this.is_array(value)) {
         for (const element of value) {
-          result.push(this.default_serializer(element, prefix + "[]"));
+          const subvalue = this.default_serializer(element, prefix + "[]");
```
jbaxendale-ut: same as the previously discussed bug

.github/workflows/ci.yml
bogdan: lets use 4.0

spec/js_routes/default_serializer_spec.rb
```diff
@@ -28,4 +29,60 @@
...
+    it "omits top-level undefined object properties" do
+      expectjs("Routes.serialize({a: 1, b: undefined})").to eq("a=1")
+    end
+
+    it "omits nested undefined object properties" do
+      expectjs("Routes.serialize({a: {b: 1, c: undefined}, d: undefined})").to eq("a%5Bb%5D=1")
+    end
+
+    it "omits empty serialized array elements" do
+      expectjs("Routes.serialize({a: [{b: undefined}, {c: 1}]})").to eq("a%5B%5D%5Bc%5D=1")
```
bogdan: As we are targeting rails compatibility, lets use `to_query` everywhere, just omit `undefined` on ruby side.

This may help with future behaviour changes.

Same in other tests in a file.

spec/spec_helper.rb
```diff
@@ -57,6 +58,7 @@ def evaljs(string, force: false, filename: 'context.js')
 end
 
 def evallib(**options)
+  options = {omit_undefined_query_parameters: false}.merge(options)
```
bogdan:
```suggestion
def evallib(omit_undefined_query_parameters: false, **options)
```

gemfiles/rails80_sprockets_4.gemfile
```diff
@@ -3,6 +3,6 @@
 source "http://rubygems.org"
 
 gem "railties", "~> 8.0"
-gem "sprockets", "~> 4.2"
+gem "sprockets", "~> 4.0"
```
jbaxendale-ut, line 6: copilot warned me of an appraisals lockfile mismatch (though it ultimately still resolves to 4.2)

spec/js_routes/omit_undefined_query_parameters_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      test_routes.thing_path(5, :optional_id => nil)
+    )
+  end
+
+  def evaluate_routes(**options)
+    evallib(module_type: nil, namespace: 'Routes', **options)
+  end
+
+  def evaluate_routes_without_default
+    JsRoutes.configuration.omit_undefined_query_parameters = nil
```
bogdan: I don't think this is legit to do as it changes config for future tests. Can we pass the config to generate below?

lib/js_routes/configuration.rb
```diff
@@ -26,6 +26,8 @@ class Configuration
     attr_accessor :camel_case
     sig { returns(Options) }
     attr_accessor :default_url_options
+    sig { returns(T.nilable(T::Boolean)) }
+    attr_accessor :include_undefined_query_parameters
```
jbaxendale-ut, line 36: I decided to flip the wording of this property, as I found some of the boolean and undefined situations hard to follow 😅

spec/js_routes/include_undefined_query_parameters_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    expectjs("Routes.thing_path(5, {optional_id: undefined})").to eq(
+      test_routes.thing_path(5, :optional_id => nil)
+    )
+  end
+
+  def evaluate_routes(**options)
+    evallib(module_type: nil, namespace: 'Routes', **options)
+  end
+
+  def evaluate_routes_without_default
```
bogdan: This is used just once, inline the method.

lib/templates/initializer.rb
```diff
@@ -22,6 +22,9 @@
   # for apps that work on multiple domains.
   # c.url_links = true
 
+  # Omit query parameters whose value is undefined.
```
bogdan, line 30:
```suggestion
  # Omit query parameters whose value is undefined.
  # This option is transitional to introduce new behaviour to old apps.
  # Recommended to stay false for new apps.
```

spec/js_routes/default_serializer_spec.rb
```diff
@@ -28,4 +29,60 @@
...
+    end
+
+    it "preserves explicit null as Rails nil" do
+      expectjs("Routes.serialize({a: null, b: {c: null, d: undefined}})").to eq(
+        {a: nil, b: {c: nil}}.to_query
+      )
+    end
+
+    it "keeps undefined array elements serialized as Rails nil" do
+      expectjs("Routes.serialize({a: [undefined]})").to eq(
```
bogdan, line 139: Lets also add test for `{a: [1, {b: undefined}, 2]}`. I belive should be equal to  `{a: [1, {}, 2]}.to_query`.

jbaxendale-ut (commented):
@bogdan I think I addressed your concerns and cleaned some things up (including changing the option config name to be less awkward to read).

Also wanted to call out something I caught doing an AI review, namely the callout about arrays getting an empty string: https://github.com/railsware/js-routes/pull/346/changes#r3319875943

So let me know what you think

bogdan (changes requested):
Like the new name 👍

spec/js_routes/default_serializer_spec.rb
jbaxendale-ut: I updated them to call `to_query` -- also FYI as a follow up to my previous long conversation with Codex over the array handling… basically, my dueling agents kind of were in conflict, but eventually I emphasized wanting to maintain matching Rails behavior than whatever the edge case was (which, seemed more correct if it perhaps left a funky situation of an empty object or string within an array possible)

> We removed the special array filtering because it made js-routes cleaner than Rails, but less Rails-compatible.
> 
>   Rails to_query does not filter empty serialized fragments inside arrays. For example:
> 
>   {a: [{}, {c: 1}]}.to_query
>   # => "&a%5B%5D%5Bc%5D=1"
> 
>   So after include_undefined_query_parameters: false omits undefined object properties, this JS shape:
> 
>   {a: [{b: undefined}, {c: 1}]}
> 
>   is equivalent to this Rails shape:
> 
>   {a: [{}, {c: 1}]}
> 
>   and Rails emits the leading &.
> 
>   The removed array guard avoided that by skipping empty sub-serializations, but that diverged from Rails’ to_query behavior. Since js-routes prioritizes matching Rails output over canonicalizing query strings, we removed the guard and added specs documenting the Rails-parity edge cases.

spec/js_routes/default_serializer_spec.rb
jbaxendale-ut: covered

spec/spec_helper.rb
jbaxendale-ut: naming change aside, flipped it to leaving the legacy behavior as the spec default

spec/js_routes/omit_undefined_query_parameters_spec.rb
jbaxendale-ut: yeah good eye, wasn't cleaning up nor necessary

spec/spec_helper.rb
```diff
@@ -56,7 +57,8 @@ def evaljs(string, force: false, filename: 'context.js')
   raise e
 end
 
-def evallib(**options)
+def evallib(include_undefined_query_parameters: true, **options)
+  options[:include_undefined_query_parameters] = include_undefined_query_parameters
```
jbaxendale-ut: handles legacy behavior as default (I think this is what you wanted right?)

jbaxendale-ut (commented):
Alrighty, called out the fixes in this round as well as a new dialog about LLM about the array behavior (tl,dr: prioritizing matching Rails behavior over "cleaner" querystring)

jbaxendale-ut:
Also did a cleanup for the PR and Squash description (thank you for this gem btw!)

spec/spec_helper.rb
bogdan: In live usage - right. It tests - I am not sure. I think we should encourage people to migrate and use future default in tests.

spec/js_routes/include_undefined_query_parameters_spec.rb
```diff
@@ -0,0 +1,62 @@
+require 'spec_helper'
+
+describe "include_undefined_query_parameters migration" do
```
bogdan, line 3: Is this now covered by `default_serializer_spec.rb?` and we don't need this file?

spec/js_routes/default_serializer_spec.rb
```diff
@@ -28,4 +29,82 @@
       ].join("\n")
     ).to eq("a=1&b=2")
```
bogdan:
```suggestion
    ).to eq({a:1, b:2}.to_query)
```

spec/js_routes/default_serializer_spec.rb
bogdan:
```suggestion
      {a: 1, b: [2,3], c: {d: 4, e: 5}, f: ''}.to_query
```

spec/js_routes/default_serializer_spec.rb
```diff
@@ -28,4 +29,82 @@
...
+    end
+
+    it "serializes mixed arrays with omitted undefined properties like Rails" do
+      expectjs("Routes.serialize({a: [1, {b: undefined}, 2]})").to eq(
+        {a: [1, {}, 2]}.to_query
+      )
+    end
+
+    it "keeps nested undefined array elements serialized as Rails nil" do
+      expectjs("Routes.serialize({a: [[undefined], {b: 1}]})").to eq(
```
bogdan, line 119: It is good you brought this one.

I made a research on how this is handled. I found 3 JS implementations of the algorithm and tested different scenarios for `undefined` handling:

Image

I had left me right it the middle as there is no consensus between these tools.

What's your opinion - how we should do it?
Do you know any other implementation of the algorithm in JS?

jbaxendale-ut:
@bogdan aiming to get back to this PR this week, a little behind and working through conflicts 😅

bogdan:
@jbaxendale-ut sorry, recent activity brought my attention to the repo and I made a radical change that planned for long time.

spec/spec_helper.rb
jbaxendale-ut: ok, I can get behind that... flipped them to "false" by default and included the "true" test cases in serializer spec

spec/spec_helper.rb
```diff
@@ -126,6 +128,8 @@ class ::App < Rails::Application
   end
 
   config.before :each do
+    JsRoutes.configuration.include_undefined_query_parameters = true
```
jbaxendale-ut, line 131: Assuming to keep this as `true` until rollout though

spec/js_routes/include_undefined_query_parameters_spec.rb
jbaxendale-ut: Removed the redundant serializer specs, kept the config ones for now

spec/js_routes/default_serializer_spec.rb
jbaxendale-ut: Hah to be honest I think I have learned far more than I ever expected about QS handling with this PR 😆 so I am not entirely sure if I am being honest.

If I had to guess, jQuery is probably the most "correct" out of them being the longest-standing project, but that is more "stood the test of time" than is "correct int he modern era".

I know that when adopting `query-string` I needed to always set https://github.com/sindresorhus/query-string#arrayformat to `bracket` to match Rails expectations... but I also would not be surprised if I've never actually encountered the extremes of these edge cases.

the `a=[object Object]` case just seems invalid though as opposed to the other two. Maybe QS is closer to the "future" behavior of omitting things. jQuery seems like it is maintaining "undefined" which is now having me question a bit of everything 😵

jbaxendale-ut: Gave Codex a whirl on which of the options were most HTML-spec compliant and the summary seems to be that Rails is (allegedly) _not_ itself strictly HTML-spec compliant:

> The comment is fair: there is no clean consensus among JS serializers because this is outside normal HTML form semantics. But if the question is most HTML-
>   spec compliant, I would not use Rails to_query as the deciding source for the empty nested-array cases.
> 
>   HTML form submission is modeled as a flat entry list of successful controls, where entries are name/value pairs. Disabled/unchecked/nameless controls are
>   skipped, and then application/x-www-form-urlencoded serializes actual tuples only, adding & only between emitted tuples. The URL serializer explicitly starts
>   with an empty output and only appends & when output is non-empty before appending the next name=value pair. Sources: HTML form entry construction and
>   conversion to name-value pairs, plus URL form-urlencoded serialization in WHATWG HTML/URL specs:
>   https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#constructing-the-entry-list
>   https://url.spec.whatwg.org/#concept-urlencoded-serializer
> 
>   So the most spec-compliant position is:
> 
>   - undefined should mean “no successful control / no entry”, so omit it.
>   - null, if intentionally supported as Rails nil, can serialize as an empty value or bare key depending on the Rails compatibility mode.
>   - Empty objects/arrays that produce no entries should not contribute empty fragments.
>   - Query strings should not contain leading &, trailing &, or doubled && from empty recursive fragments.
> 
>   That means HTML-spec-compliant output would prefer:
> 
>   Routes.serialize({a: [{b: undefined}, {c: 1}]})
>   # => "a%5B%5D%5Bc%5D=1"
>   # not "&a%5B%5D%5Bc%5D=1"
> 
>   and:
> 
>   Routes.serialize({a: [1, {b: undefined}, 2]})
>   # => "a%5B%5D=1&a%5B%5D=2"
>   # not "a%5B%5D=1&&a%5B%5D=2"
> 
>   The caveat: Rails compatibility and HTML-spec compliance diverge here. Rails to_query preserves those empty array fragments and can emit leading/double
>   separators. If js-routes’ priority is “byte-for-byte Rails to_query compatibility,” then the current Rails-parity specs are defensible. If the goal is
>   “future behavior aligned with HTML/URL form encoding,” then filtering empty recursive serializations is the more spec-compliant choice.

jbaxendale-ut: In regard to the three libraries, it disagrees with my assumptions about jQuery (which maybe is targeting max compatibility… presumably like Rails, rather than spec compliance?):

> Of the three in the screenshot, I’d call query-string the most HTML-spec compliant overall, with a caveat.
> 
>   Why:
> 
>   - HTML form submission serializes a flat ordered list of successful name=value entries.
>   - Missing/non-successful controls simply do not produce entries.
>   - Repeated values are naturally represented as repeated same-name entries, like a=1&a=2.
>   - The form-urlencoded serializer joins emitted entries with &; it does not emit placeholder separators for omitted values.
> 
>   That maps closest to query-string in the table:
> 
>   {a: undefined}          -> omitted
>   {a: undefined, b: 1}    -> b=1
>   {a: [undefined]}        -> omitted
>   {a: [1, undefined]}     -> a=1
>   {a: [1, undefined, 2]}  -> a=1&a=2
> 
>   qs is also spec-aligned on omitting undefined, but its indexed output:
> 
>   a[0]=1&a[2]=2
> 
>   is less HTML-form-like than a=1&a=2. HTML does not have array indexes as a form encoding concept; it just has multiple controls with the same name.
> 
>   jQuery.param is the least spec-compliant for this specific question because it treats undefined as an emitted empty value:
> 
>   {a: undefined} -> a=
> 
>   That looks more like “a successful control with an empty value” than “no value/property existed,” which is not what undefined usually means.
> 
>   Caveat: for nested objects, none of the three is purely HTML-spec compliant because HTML form data is flat. But among these behaviors, query-string is the
>   closest to the HTML/URL model for undefined omission and repeated values. 

Notably calling out the [object Object] being kind of hosed, so I'm not sure where this lands us beyond "omitting undefined items _does_ seem correct" on top of maybe we should be recursing more but OK to avoid emptying that off as I think an empty but defined item is more legitimate than something that never got allocated to begin with?

jbaxendale-ut (commented):
Ok I think I've resolved the merge conflicts and got this back to working, although not sure that I have the best answer for the "how should this case work" as I am somewhat as a loss for words on the library behaviors (though talking with Codex gives me a bit more confidence in this PR in the end?)

spec/js_routes/rails_routes_compatibility_spec.rb
```diff
@@ -215,6 +215,12 @@
     end
 
     it "should support null and undefined parameters" do
+      expectjs("Routes.inboxes_path({uri: null, key: undefined})").to eq(test_routes.inboxes_path(:uri => nil))
+    end
+
+    it "should support legacy null and undefined parameters when enabled" do
```
bogdan: This one now need to go to `options_spec`

spec/js_routes/default_serializer_spec.rb
bogdan: Rails core team was surprisingly welcome to fix unbalanced `&`: https://github.com/rails/rails/pull/57500.

We should not be compliant with Rails bugs like this. Previously there were issues like this with `false` in route parameters and I didn't bother implementing unfixed rails behavior for previous versions and was just marking test pending for `Rails.version < x.x`. 

https://github.com/rails/rails/pull/42283

If you know any other bug with it, send it to rails or let me know if you need any help.

We might have some issues testing compatibility with fixed but unreleased rails serializer bugs. I believe that would require testing again rails head, which I think is fine.

Obviously, rails would not care about `undefined` though.
And I didn't find any description in spec neither.

spec/spec_helper.rb
```diff
@@ -56,7 +57,8 @@ def evaljs(string, force: false, filename: 'context.js')
   raise e
 end
 
-def evallib(**options)
+def evallib(include_undefined_query_parameters: false, **options)
+  options[:include_undefined_query_parameters] = include_undefined_query_parameters
```
bogdan: Simplify:

```suggestion
def evallib(**options)
  options[:include_undefined_query_parameters] ||= false 
```

bogdan (commented):
Getting close.

spec/spec_helper.rb
```diff
@@ -57,6 +58,7 @@ def evaljs(string, force: false, filename: 'context.js')
 end
 
 def evallib(**options)
+  options[:include_undefined_query_parameters] ||= false
```
jbaxendale-ut, line 61: fixed this up

gemfiles/rails80_sprockets_4.gemfile
```diff
@@ -2,7 +2,7 @@
 
 source "http://rubygems.org"
 
-gem "railties", "~> 8.0"
-gem "sprockets", "~> 4.2"
+gem "railties", "~> 8.0.0"
```
jbaxendale-ut, line 5: fixed these being too loose to resolve properly

spec/js_routes/rails_routes_compatibility_spec.rb
```diff
@@ -214,8 +214,8 @@
       )
     end
 
-    it "should support null and undefined parameters" do
-      expectjs("Routes.inboxes_path({uri: null, key: undefined})").to eq(test_routes.inboxes_path(:uri => nil, :key => nil))
+    it "should support null and omit undefined parameters" do
```
jbaxendale-ut, line 217: tweaked this wording a bit

spec/js_routes/default_serializer_spec.rb
```diff
@@ -26,6 +27,119 @@
...
+    end
+
+    it "omits nested undefined object properties" do
+      expectjs("Routes.serialize({a: {b: 1, c: undefined}, d: undefined})").to eq(
+        {a: {b: 1}}.to_query
+      )
+    end
+
+    it "serializes array objects with omitted undefined properties without empty separators" do
+      pending("Rails main fixed empty hash array serialization in rails/rails#57500")
```
jbaxendale-ut, line 103: I marked these as pending at the moment, as they are the "right" case but we are still resolving to the released Rails 8.1.3 which is with the bug... I am not 100% sure how you prefer to handle this at the moment, so I opened another PR against this on my fork with some options if you could take a peek?

https://github.com/user-interviews/js-routes/pull/3/changes

spec/js_routes/default_serializer_spec.rb
jbaxendale-ut: I put up a draft PR in my fork repo if you could take a look at some of the options, I think I addressed the issue but I'm not totally sure if I'm handling the specs and busted prior versions the way you would like:

https://github.com/user-interviews/js-routes/pull/3#pullrequestreview-4439415113

jbaxendale-ut (commented):
I think I addressed the small bits, but we still have to make some calls on the rails bugfix

PR #347, opened 2026-06-02
Add configurable output path to middleware generator

alec-c4:
## Summary

- The middleware generator now accepts an optional argument for the output location (directory or full file path)
- The `file` config option now accepts a **directory** (e.g. `"app/frontend"`) in addition to a full path — the default filename is appended automatically
- `FileUtils.mkdir_p` added before file writes so the output directory is created automatically if it does not exist

## Usage

```sh
# Default — app/javascript/routes.js
rails generate js_routes:middleware

# Custom directory — creates app/frontend/routes.js + routes.d.ts
rails generate js_routes:middleware app/frontend

# Custom full path
rails generate js_routes:middleware app/frontend/my_routes.js
```

The generator writes `c.file` into the initializer and adds the correct paths to `.gitignore` automatically.

The same directory syntax works in the initializer:

```ruby
JsRoutes.setup do |c|
  c.file = "app/frontend"          # => app/frontend/routes.js
  # or
  c.file = "app/frontend/routes.js"
end
```

## Test plan

- [ ] `rails generate js_routes:middleware` — generates at default path, no `c.file` in initializer
- [ ] `rails generate js_routes:middleware app/frontend` — creates `app/frontend/routes.js` + `.d.ts`, initializer has `c.file = "app/frontend"`, `.gitignore` shows `/app/frontend/routes.js` and `/app/frontend/routes.d.ts`
- [ ] `rails generate js_routes:middleware app/frontend/my_routes.js` — full path variant works end-to-end
- [ ] `rake js:routes` with a non-existent output directory creates the directory automatically
- [ ] 245 RSpec examples pass (1 pre-existing failure in `dts_spec` due to missing `node_modules` in CI env, unrelated to this PR)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -53,7 +53,14 @@ to automatically generate and maintain `routes.js` file and corresponding
 Run a command:
 
 ``` sh
+# Default location (app/javascript/routes.js)
 rails generate js_routes:middleware
+
+# Custom output directory — generates app/frontend/routes.js
+rails generate js_routes:middleware app/frontend
+
+# Custom full path
+rails generate js_routes:middleware app/frontend/my_routes.js
 ```
 
 #### Setup Manually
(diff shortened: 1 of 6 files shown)


bogdan:
There've been a recent commit:
https://github.com/railsware/js-routes/commit/10c831ba2cceaba113c07a39acdf65d147ee81c6

Do you think support of custom location is needed when it can take configuration from rails?

alec-c4:
@bogdan yep, because it doesn't work with vite based configurations, such as inertia-rails

bogdan:
Can you explain?

alec-c4:
@bogdan sure :) There are several configurations, based on https://vite-ruby.netlify.app/guide/rails.html gem - plain vite_rails and [inertia_rails](https://inertia-rails.dev). They use `app/frontend` directory for frontend stuff, but `js-routes` do not identify this location. I've tried to put initializer with correct path before I run `rails generate js_routes:middleware`, generator asks me to overwrite initializer, puts routes.js file to the correct directory, but adds to `.gitignore` wrong `app/javascript` path, so I decided to fix this issue completely :)

bogdan:
> They use app/frontend directory for frontend stuff, but js-routes do not identify this location

Did you configure `config.javascript_path`? I think you should do it, so that rails side knows the right location.

>  but adds to .gitignore wrong app/javascript path

I think this is what should be fixed instead of broader issue. 

See, generator is designed for the most trivial setup. It gives you no options to make sure you have a quick start. You are in situation where quick start is not an option because you have custom setup.

PR #350, opened 2026-06-08
Avoid empty query separators for Rails-head arrays

jbaxendale-ut:
## Summary

This updates the default serializer to avoid producing malformed query separators when arrays contain values that serialize to an empty query fragment.

Rails fixed the corresponding `Array#to_query` behavior in rails/rails#57500. Before that fix, empty hashes inside arrays can leave leading, trailing, or doubled `&` separators, for example:

```ruby
{a: [1, {}, 2]}.to_query
# older Rails: "a%5B%5D=1&&a%5B%5D=2"
# fixed Rails: "a%5B%5D=1&a%5B%5D=2"
```

js-routes can hit the same shape directly with empty objects, or indirectly when `undefined` properties are omitted and an object becomes empty.

## Changes

- Skip empty serialized array entries so arrays do not produce leading, trailing, or doubled separators.
- Preserve Rails-compatible nested empty-array behavior, where an empty array inside another array serializes as a nil query value.
- Keep serializer expectations based on Rails `to_query`; for Rails versions before the upstream fix, specs normalize only the known separator bug.

## Coverage

Adds/updates coverage for:

- empty objects in the middle, beginning, and end of arrays
- nested empty objects inside arrays
- omitted `undefined` properties that leave empty objects inside arrays
- nested empty arrays, to ensure that behavior is unchanged
- custom serializer filtering that previously produced a trailing `&&`

### DIFF
--- a/spec/js_routes/default_serializer_spec.rb
+++ b/spec/js_routes/default_serializer_spec.rb
@@ -13,6 +13,20 @@ describe JsRoutes, "#serialize" do
     evallib(**options)
   end
 
+  def rails_fixed_empty_hash_array_query?
+    ActiveSupport.gem_version >= Gem::Version.new("8.2.0.alpha")
+  end
+
+  def fixed_empty_hash_array_query(value)
+    result = value.to_query
+    return result if rails_fixed_empty_hash_array_query?
+
+    # Rails versions before this fix can emit leading, trailing, or doubled
+    # separators for empty hashes inside arrays. Keep deriving expectations from
+    # to_query, then normalize only that known separator bug.
+    result.gsub(/\A&+|&+\z|(&){2,}/) { $1 ? '&' : '' }
+  end
+
   it "serializes basic object, array, nested object, and empty string values like Rails" do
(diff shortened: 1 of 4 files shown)


jbaxendale-ut:
Just noting that this is the version of the PR without the `CGI.escape` handling, which I dug into more and explained a bit https://github.com/user-interviews/js-routes/pull/3#discussion_r3376674818

spec/js_routes/default_serializer_spec.rb
```diff
@@ -13,6 +13,20 @@
     evallib(**options)
   end
 
+  def rails_fixed_empty_hash_array_query?
+    ActiveSupport.gem_version >= Gem::Version.new("8.2.0.alpha")
```
jbaxendale-ut, line 17: Wasn't sure if they were going to include this fix in `8.1.4` but it seems like it might be different enough that it only stays in the `8.2.x` line

bogdan: Normally they don't backport bugs unless explicitly asked. So, unlikely it will get into `8.1.x`.

